PR #11, opened 2016-03-30
Use binded fields instead of finding them by id again

DmitriyZaitsev:
We can get views from Binding directly. It's much faster than using findViewById and we don't have to do extra work.


JoseAlcerreca:
LGTM thanks!

PR #14, opened 2016-03-30
Fixes issue with data not being stored in local data source

JoseAlcerreca:
@ErikHellman @malmstein some fixes to the repo that need to be merged into child projects. PTAL.


ErikHellman:
LGTM!

PR #26, opened 2016-04-04
Custom Guava module

malmstein:
#### Problem

Guava is a very helpful library, however it's a huge dependency that adds 13K methods.

#### Solution

There is now a custom module in the project with only the packages that we are using. We are using ProGuard to only keep the dependencies that are needed.

#### Fixes

https://github.com/googlesamples/android-architecture/issues/22
https://github.com/googlesamples/android-architecture/issues/24


todoapp/guava_custom/.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
JoseAlcerreca: should be unnecessary.

JoseAlcerreca:
As agreed offline, we're going back to proguard and the new shrinker. It makes the build faster using guava reducing method count from 34413 to 10118. The custom module is a better solution but distracting for a sample.

PR #30, opened 2016-04-05
Prevent NullPointerException in ActionBar

Rainer-Lang:


Rainer-Lang:
I signed it!

JoseAlcerreca:
If you have seen this happening please specify in which circumstances. Also, silently checking for this null and not throwing an error or logging doesn't seem like a good idea to me.

JoseAlcerreca:
Ping!

Rainer-Lang:
@JoseAlcerreca I didn't "see" it in your samples. I think NPE should be prevented whereever possible.
You're right only a silent check isn't good - especially for a "blueprint" like these samples.
This NPE could happen (maybe not here but if the blueprint is used from devs) and I think we should care about this fact.

JoseAlcerreca:
Since you agree on the disadvantages of the silent check, I'll close this PR now. Thanks!

PR #36, opened 2016-04-07
fix tools context in layout

quangson91:


JoseAlcerreca:
Thanks! Next time simply modify todo-mvp. It's better to merge changes from it than repeating the same change over all variants.

quangson91:
Ok, Thank you.

PR #48, opened 2016-04-20
Fix crash in support design library(v.23.2.+)

h3r3x3:
Fix "View can not be anchored to the the parent CoordinatorLayout" error in new version support design library(v.23.2.+)

### DIFF
--- a/todoapp/build.gradle
+++ b/todoapp/build.gradle
@@ -23,7 +23,7 @@ task clean(type: Delete) {
 // Define versions in a single place
 ext {
     // Sdk and tools
-    minSdkVersion = 14
+    minSdkVersion = 10
     targetSdkVersion = 22
     compileSdkVersion = 23
     buildToolsVersion = '23.0.2'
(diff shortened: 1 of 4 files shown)


todoapp/build.gradle
```diff
@@ -23,13 +23,13 @@ task clean(type: Delete) {
 // Define versions in a single place
 ext {
     // Sdk and tools
-    minSdkVersion = 10
+    minSdkVersion = 14
```
malmstein: @JoseAlcerreca do we care about this?

h3r3x3: Wow, it's my mistake (×_×)

JoseAlcerreca: Yes we do.

JoseAlcerreca:
Sorry I didn't approve this PR earlier. There's a problem with the FAB not going back down after a Snackbar is shown, but it turns out it's a bug in the support lib. This happens only when animations are disabled, so this PR LGTM.

PR #67, opened 2016-05-09
data binding is not in beta anymore (since Android Gradle Plugin 1.5.0)

yigit:


yigit:
I signed it!

yigit:
I replied w/ `I signed it!` assuming it is some automated flow :) Meanwhile, I work at Google so I don't need an additional CLA.

JoseAlcerreca:
It is automated :) but multiple accounts and two different CLAs makes it difficult. 

Thanks!

PR #90, opened 2016-05-31
Added null check in GetTask.

NikitaKozlov:
Removed null check from TaskDetailPresenter.

[https://github.com/googlesamples/android-architecture/issues/65](url)


flipper83:
👍 thank you! This change has sense.

PR #112, opened 2016-07-04
private constructor to prevent instantiating

jydimir:
To prevent accidental instantiating the contract class constructor must be private.


jydimir:
@googlebot 
I signed it!

malmstein:
good catch! 👍

Issue #110, opened 2016-07-01
@NonNull annotation and checkNotNull() used incorrectly


samiuelson:
I noticed very incorrect usage of `checkNotNull()` guava method for objects with `@NonNull` annotations. 
Example from TasksPresenter class:

```
@Override
    public void openTaskDetails(@NonNull Task requestedTask) {
        checkNotNull(requestedTask, "requestedTask cannot be null!");
        mTasksView.showTaskDetailsUi(requestedTask.getId());
    }

    @Override
    public void completeTask(@NonNull Task completedTask) {
        checkNotNull(completedTask, "completedTask cannot be null!");
        mTasksRepository.completeTask(completedTask);
        mTasksView.showTaskMarkedComplete();
        loadTasks(false, false);
    }

    @Override
    public void activateTask(@NonNull Task activeTask) {
        checkNotNull(activeTask, "activeTask cannot be null!");
        mTasksRepository.activateTask(activeTask);
        mTasksView.showTaskMarkedActive();
        loadTasks(false, false);
    }
```
Guys, read the docs please. `@NonNull` is used to assure that related object is never ever null. The `checkNotNull()` is used incorrectly therefore. `checkNotNull()` could be called on particular object before passing it with some of these methods above, by method call.
It's only one example. Such type of mistake is common among whole project. It's not a bug but it's very misleading and ugly.

Rainer-Lang:
Do you want to make a pr?

samiuelson:
Yes, I can fix it.

JoseAlcerreca:
As agreed in #113 , closing!

PR #113, opened 2016-07-05
Remove incorrectly used checkNotNull() method calls

samiuelson:
If some object has @NonNull annotation it tells it can't be null. `checkNotNull()` calls with that annotation are redundant and misleading.  #110


k0ber:
Actually `@NonNull` is used to show that you don't expect null parameter, but anyone can pass it to your method, Android studio will just show a warning. Maybe it happens not intentionally, but it can happen. To check that everything is ok **in Runtime**, we can use  `checkNotNull()`.  I think, it's right to use them both.

jydimir:
`@NonNull` is for _static analysis_(for compiler and static analyzer tools like FindBugs and other). `checkNotNull()` is for _runtime_ checks(as @k0ber wrote). So this is not redundant.

JoseAlcerreca:
Yep as k0ber and jydimir say, these checks are needed to enforce at runtime. Thanks for the PR though!

PR #119, opened 2016-07-18
Cache getTask() result

Rainer-Lang:
In ```TasksRepository``` a task is loaded with ```mTasksRemoteDataSource.getTask()``` and ```mTasksLocalDataSource.getTask()``` - but the task isn't cached.
With this PR the loaded task is cached in the provided map ```mCachedTasks```.

I noticed this in the branches todo-mvp and dev-todo-mvp-clean-rxjava (@NikitaKozlov).

### DIFF
--- a/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/TasksRepository.java
+++ b/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/TasksRepository.java
@@ -221,6 +221,7 @@ public class TasksRepository implements TasksDataSource {
         mTasksLocalDataSource.getTask(taskId, new GetTaskCallback() {
             @Override
             public void onTaskLoaded(Task task) {
+                mCachedTasks.put(task.getId(), task);
                 callback.onTaskLoaded(task);
             }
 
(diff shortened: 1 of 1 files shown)


malmstein:
Hi @Rainer-Lang, could you add a description to the PR? Explaining the problem and the solution you are following.

Rainer-Lang:
@malmstein Description updated.
Thanks for the feedback.

JoseAlcerreca:
The idea sounds good to me! `mCachedTasks` might be null though,

Rainer-Lang:
@JoseAlcerreca Should I add ```checkNotNull(mCachedTasks);```?

JoseAlcerreca:
No, the rest of the methods create the map if it doesn't exist. But we need to refactor it to avoid repetition.

Rainer-Lang:
```deleteTask()``` with ```mCachedTasks.remove(taskId);``` could be a problem.

JoseAlcerreca:
@Rainer-Lang do you want to have a go at it? For now simply create the mCachedTasks if it's null, like the other methods do.

Rainer-Lang:
@JoseAlcerreca Please have a look. Thanks Jose.

JoseAlcerreca:
Thanks @Rainer-Lang!

PR #148, opened 2016-08-18
AddEditTaskPresenter - implementation and tests finished

florina-muntenescu:


todoapp/app/src/test/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskPresenterTest.java
```diff
@@ -58,14 +57,17 @@ public void setupMocksAndView() {
         // inject the mocks in the test the initMocks method needs to be called.
         MockitoAnnotations.initMocks(this);
 
+        mSchedulerProvider = new ImmediateSchedulerProvider();
```
startthread: For testing, I think it would be better if schedulers are provided through `@Rule` and RxJava hooks.
Should create TestRule and use rule in AddEditTaskPresenterTest.

```
public class RxJavaResetRule implements TestRule {

    @Override
    public Statement apply(final Statement base, Description description) {

        return new Statement() {
            @Override
            public void evaluate() throws Throwable {

                RxJavaHooks.setOnComputationScheduler(new Func1<Scheduler, Scheduler>() {
                    @Override
                    public Scheduler call(Scheduler scheduler) {
                        return Schedulers.immediate();
                    }
                });
                

                base.evaluate();

                RxJavaHooks.setOnComputationScheduler(null);
            }
        };
    }
}```

PR #166, opened 2016-09-01
name change

florina-muntenescu:


JoseAlcerreca:
This was done totally consciously so that contributors get into the list in `master` 😆

PR #174, opened 2016-09-05
Instantiate the ScheduleProvider at the same time the INSTANCE is declared

thagikura:
Since the current version of code doesn't look thread safe and I thought a null may be returned in the getInstance method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,12 @@
 # TODO-MVP-RXJAVA
 
-It's based on the TODO-MVP sample and uses RxJava for communication between the data model and presenter layers.
+Project owners: [Erik Hellman](https://github.com/erikhellman) & [Florina Muntenescu (upday)](https://github.com/florina-muntenescu)
 
 ### Summary
 
-Compared to the TODO-MVP, both the Presenter contracts and the implementation of the Views stay the same. The changes are done to the data model layer and in the implementation of the Presenters.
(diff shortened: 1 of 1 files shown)


JoseAlcerreca:
What about synchronizing the getInstance() method instead? The whole point is to lazy-load the scheduler I think.

thagikura:
Yes, synchronizing the getInstance method also works. Okay, I'll change the PR to do so.
(Sorry looks like I accidentally closed this PR. I may send another PR if I can't find a way to amend the PR)

thagikura:
Sent another PR #178 to make the method synchronized.

NikitaKozlov:
@thagikura why do you want to keep lazy initialisation at all? Whenever somebody access `SchedulerProvider`, the only thing he can do is call `getInstance()`. So you can just put initialisation in the field itself, put `volatile` and make it `final`.

thagikura:
@NikitaKozlov I'm fine if the cod follows lazy initialization or not as long as thread safety is guaranteed.
Actually the first PR was instantiating the ScheduleProviders in the field itself.

So I want to follow the owner's preference regarding that point ( @JoseAlcerreca or @florina-muntenescu ?)

NikitaKozlov:
@thagikura My point is, that even if you put initialization in the field itself, instance creation will happened only on first `getInstance()`call,  since there is no other way to interact with this class.
But it is true, it is up to owners to decide.

PR #177, opened 2016-09-07
Update buildTools, supportLibrary and classpath version.

filipedelimabrito:


todoapp/build.gradle
```diff
@@ -25,11 +25,11 @@ ext {
     // Sdk and tools
     minSdkVersion = 10
     targetSdkVersion = 22
-    compileSdkVersion = 23
-    buildToolsVersion = '23.0.2'
+    compileSdkVersion = 24
+    buildToolsVersion = '24.0.1'
```
Rainer-Lang: You could use ```buildToolsVersion = '24.0.2'```

filipedelimabrito: You are right! Thanks!

JoseAlcerreca:
Thanks for the PR! but we need to do this from `todo-mvp` and then merge into the branches.

Issue #154, opened 2016-08-23
I need advice. branch/todo-mvp-clean. LongRunningUseCase


westman2000:
Suppose there is such [UseCase](https://github.com/westman2000/MVPCAndroid/blob/mvpc-loaders/app/src/main/java/corp/wmsoft/android/examples/mvpc/longrunning/iteractor/LongRunningUseCase.java)

And [Presenter](https://github.com/westman2000/MVPCAndroid/blob/mvpc-loaders/app/src/main/java/corp/wmsoft/android/examples/mvpc/longrunning/LongRunningPresenter.java) who using it.

When user start UseCase, i show loading, but because this is long running operation, user can close this screen(press back button, close activity), or even close application(press home button).
In this case, will be called detachView() and onDestroyed() ([presenter will be destroyed here](https://github.com/westman2000/MVPCAndroid/blob/mvpc-loaders/mvpcandroid/src/main/java/corp/wmsoft/android/lib/mvpc/presenter/loader/MVPCPresenterLoader.java))

But since the operation is continued in the background, presenter will receive 
`onSuccess(LongRunningUseCase.ResponseValue response)`
but because view already detached, there will be Error.

And my questions are:
1. How to avoid this error? I understand i can just check if view is attached `isViewAttached()`. but maybe there is a more elegant solution? For example how I can unsubscribe from UseCase if presenter destroyed, and then callback will not be called?
2. I do not understand how and why there is a function call `onSuccess` in the presenter, if presenter was already destroyed in `MVPCPresenterLoader.onReset()`
3. Somebody thought to create an intermediate buffer(let's called ViewState) between the Presenter and View to save commands for the View. In case you Presenter was disconnected from the View, and background thread finish loading, we need to hide the ProgressBar and display data. But this commands must be send to View only after attachView... In generally i speak about somesing like [this](https://github.com/Arello-Mobile/Moxy/wiki)

ErikHellman:
The solution here is to have the `UseCaseHandler` return an object that implements something like `cancel()`. The presenter needs to define a `stop()` or `unsubscribe()` method that the view will call once it gets detached.

Issue #155, opened 2016-08-24
Make it possible to cancel a running use case in todo-mvp-clean


ErikHellman:
`UseCaseHandler.execute()` needs to return an object that lets us cancel an ongoing task. This is to prevent the callback from leaking a reference to the view. The cancel operation doesn't need to abort the thread, but should basically clear the reference to the callback to avoid the leak. See #154

gokhanbarisaker:
I solved that by returning a Subscription(direct copy from RxJava) which UseCase implements. e.g.,

```
Subscription subscription = UseCaseHandler.execute(doFoo, ...);
```
where `UseCaseHandler.execute()` returns `UseCase` itself internally.

flipper83:
I've reviewed the code and we are leaking the callback reference, my fault. I don't like to create a cancel method for avoiding this one because is move the decision to the developer and it's easy to forget it. My idea is to take weak references and not a strong reference for the callback, this will and auto cancel for those calls that don't have a callback attached. 

I will add a pull request during this weekend.

gokhanbarisaker:
@flipper83 isn't it risky to trust object to be garbage collected from memory to unsubscribe from callbacks? and most likely leak-prune

ErikHellman:
Yeah, have to agree with @gokhanbarisaker. A weak reference is not very optimal. We already have subscribe/unsubscribe in other samples, so it should be fine to use that here as well. The view tells the presenter when it is being detached.

tunjid:
Closing as obsolete in the context of the [current architecture guide](https://developer.android.com/topic/architecture).

PR #183, opened 2016-09-10
Add cancellation support to running UseCase

gokhanbarisaker:
This commits clones RxJava 1.x subscription logic in its most basic form.

Fix: #154, #155

### DIFF
--- a/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/BasePresenter.java
+++ b/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/BasePresenter.java
@@ -20,4 +20,5 @@ public interface BasePresenter {
 
     void start();
 
+    void stop();
 }
--- a/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/Subscription.java
+++ b/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/Subscription.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2016, The Android Open Source Project
+ *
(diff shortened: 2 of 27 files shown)


gokhanbarisaker:
Removed a personal TODO note as a fixup at `c65a66d`. Squashed commit with fixup to `c7cfe75` and updated pull request

gokhanbarisaker:
hi @flipper83. Is it possible for you to review this?

flipper83:
@gokhanbarisaker yes, I will try to review between this night and tomorrow morning.

flipper83:
@gokhanbarisaker sorry for delay in my responses but I had been really full of work. I going to review right now and I will send you some comments

flipper83:
After reviewing the code, some comments about this approach: For me this is overengineering the problem, you only create a subscription because you can have a memory leak in the uses cases, but you leak the architecture moving this problem to the user domain. This is not good because thread management is an implementation detail and we can fix it with another approach.

This problem doesn't need a reactive approach because you only have one response to each call, don't need a subscription that only has one response, it increases the difficulty of the domain calls, adding features that you don't use. Using an approach similar a Promises could be a better option.

On moving the code from inheritance to interfaces. You can create a lot of duplicate code, and move some dependencies from constructor to method, this will create problems with testing, or try to move it to DI approach. 

I going to write an example for fix the problem without move the cancelled to the user, and I show you.

JoseAlcerreca:
@flipper83 can we close this?

flipper83:
No yet please, I'm finishing an alternative, and I want write here the response. Sorry.

JoseAlcerreca:
Great!

PR #185, opened 2016-09-11
Fix configuration change bug when editing an existing task.

gregpearce:
The taskId was being lost during a configuration change causing the task to be saved as a new task rather than editing the existing task.


JoseAlcerreca:
Oh wow, good catch. We should have a UI test for that too. Interested?

JoseAlcerreca:
The test was a bit tricky so I created it here. 
https://github.com/JoseAlcerreca/android-architecture/pull/3

PR #189, opened 2016-09-13
fix to take care of the order of method calls

sfuku7:
Hi,

Could you please review this ?

This patch makes the unit test method calls order sensitive.

I guess the transition from 'true' to 'false' is important for the test.
mTasksView).setLoadingIndicator(true);
mTasksView).setLoadingIndicator(false);

### DIFF
--- a/todoapp/app/src/test/java/com/example/android/architecture/blueprints/todoapp/taskdetail/TaskDetailPresenterTest.java
+++ b/todoapp/app/src/test/java/com/example/android/architecture/blueprints/todoapp/taskdetail/TaskDetailPresenterTest.java
@@ -24,10 +24,12 @@ import org.junit.Before;
 import org.junit.Test;
 import org.mockito.ArgumentCaptor;
 import org.mockito.Captor;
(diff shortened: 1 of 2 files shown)


sfuku7:
I signed it!

JoseAlcerreca:
Hi! thanks for the PR! It looks good, is this the only place where this is needed in todo-mvp?

sfuku7:
Thank you for the review !

Regarding under the java logic tests directory, 'todoapp/app/src/test',
I could find the same situation on the another test code: 'TaskDetailPresenterTest'.
($ find todoapp/app/src/test -type f | xargs grep setLoadingIndicator)

Should I create another commit for that on this PR ?
Or should I create another PR for the 'TaskDetailPresenterTest' ?

JoseAlcerreca:
Another commit on this PR is fine, thanks!

sfuku7:
Thanks ! I've created another commit !

JoseAlcerreca:
Thanks!

PR #197, opened 2016-09-22
refactor: Starting loader for showing tasks earlier

DmitryMalkovich:
Load data from db and only after that, update the list with data from server


DmitryMalkovich:
I signed it!

DmitryMalkovich:
Please, check again, I signed

PR #210, opened 2016-10-11
Adding CircleCI configuration file circle.yml

### LINKED ISSUE #209: [all] adding CircleCI configuration file
At the moment there is only `TravisCI` configuration file. Adding those for `CircleCI` would be fairly easy to do and Jose said it could be a nice addition to the project.
I am willing to do it if welcomed. In the case I need to merge on all projects, do I need to create a PR for each or one would be enough for the merge-master to pass it on the others?

oldergod:
fix #209 
### Cache
- CircleCI will [automatically cache `.m2` and `.gradle` folders](https://circleci.com/docs/how-cache-works/) so did not include them in the config file.
- Also, CircleCI does not cache folder after the tests are run but [before](https://discuss.circleci.com/t/gradle-build-phase/1287/3). Hence the addition of the `./gradlew dependencies` step to allow the dependencies to actually get cached.
### Test Reports
- CircleCI supports it so I save the reports in the corresponding folder.
### Misc
- I use `cd todoapp;./gradlew xxx` oneliner because directory changes are reset on every step.
- Not sure why there was the `rm -f ~/.gradle/caches/modules-2/modules-2.lock` line but copied it in case.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,16 @@
+machine:
+  java:
+    version: oraclejdk8
+
+dependencies:
+  pre:
+    - echo y | android update sdk --no-ui --all --filter "tools,platform-tools,build-tools-24.0.2,android-24,extra-android-m2repository"
+  override:
+    - cd todoapp;./gradlew dependencies
+  post:
+    - rm -f ~/.gradle/caches/modules-2/modules-2.lock
+
+test:
+  override:
+    - cd todoapp;./gradlew test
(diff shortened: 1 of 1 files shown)


JoseAlcerreca:
Probably a dumb question: how do we know it works? There's no PR on your repo so it didn't kick in.

oldergod:
@JoseAlcerreca You can see the builds I had on my fork here: https://circleci.com/gh/oldergod/android-architecture

JoseAlcerreca:
Great thanks, I'll merge it when we get Firebase Test Lab access.

PR #220, opened 2016-10-20
Update ToDoApplication.java

mladenrakonjac:
We do not need  .tasksRepositoryModule(new TasksRepositoryModule())  cause it does not require external state.


malmstein:
Thanks @mladenrakonjac! @saulmm what do you think?

saulmm:
looks good to me! 👍

JoseAlcerreca:
Thanks!

Issue #217, opened 2016-10-17
[todo-mvp] Bug: EditTask doesn't save state on a configuration change


passsy:
When editing a task (changing the title), the changes do not survive a configuration change.

Reason: `start()` is called from the Fragments `#onResume()`. This triggers `populateTask()` resetting the UI to the state of the loaded task from the repository. The Presenter isn't aware of changes of the TextViews in the Fragment.

This happens in two cases:
- configuration change
- bring Activity to foreground `#onResume()`

Here is a test highlighting this problem. It works but needs some adjustments

``` java
    @Test
    public void configurationChange_retainsState() throws Throwable {

        // TOOD check for correct context. targetContext isn't correct
        final TasksRepository tasksRepository = Injection
                .provideTasksRepository(InstrumentationRegistry.getTargetContext());
        final String taskId = "0001";
        final Task task = new Task("Title", "description", taskId);
        tasksRepository.saveTask(task);

        // TODO is there a nicer way to start a activity with a IntentsTestRule and Intent arguments
        final Intent intent = new Intent(InstrumentationRegistry.getTargetContext(),
                AddEditTaskActivity.class);
        intent.putExtra(AddEditTaskFragment.ARGUMENT_EDIT_TASK_ID, "0001");
        mAddTaskIntentsTestRule.launchActivity(intent);

        onView(withId(R.id.add_task_title)).check(matches(withText("Title")));
        onView(withId(R.id.add_task_description)).check(matches(withText("description")));

        onView(withId(R.id.add_task_title)).perform(replaceText("Foo"));
        onView(withId(R.id.add_task_description)).perform(replaceText("Bar"));

        mAddTaskIntentsTestRule.runOnUiThread(new Runnable() {
            @Override
            public void run() {
                // TODO handle recreate for API 10
                mAddTaskIntentsTestRule.getActivity().recreate();
            }
        });

        onView(withId(R.id.add_task_title)).check(matches(withText("Foo")));
        onView(withId(R.id.add_task_description)).check(matches(withText("Bar")));
    }
```

savepopulation:
I think all samples (exclude loader implementation because loaders can survive orientation change) cannot survive orientation change. For example when you rotate the phone while getting tasks, app re send same request in mvp clean architecture and rxjava sample. If i'm wrong please correct me.

passsy:
IMHO handling a configuration change should be the minimal requirement for any sample implementations which claim to be better than putting all code into an Activity. 
[...]

PR #226, opened 2016-11-01
AddEditTask: adds rotation tests and fixes persistance bug

JoseAlcerreca:
Fixing  #217. 

setRetainInstance was not needed, turns out. Not super happy with the `shouldLoadDataFromRepo`. Suggestions?

### DIFF
--- a/todoapp/app/src/androidTest/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksScreenTest.java
+++ b/todoapp/app/src/androidTest/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksScreenTest.java
@@ -16,31 +16,8 @@
 
 package com.example.android.architecture.blueprints.todoapp.tasks;
 
-import static android.support.test.InstrumentationRegistry.getTargetContext;
-import static android.support.test.espresso.Espresso.onView;
-import static android.support.test.espresso.Espresso.openActionBarOverflowOrOptionsMenu;
-import static android.support.test.espresso.action.ViewActions.click;
-import static android.support.test.espresso.action.ViewActions.closeSoftKeyboard;
-import static android.support.test.espresso.action.ViewActions.replaceText;
-import static android.support.test.espresso.action.ViewActions.typeText;
-import static android.support.test.espresso.assertion.ViewAssertions.doesNotExist;
-import static android.support.test.espresso.assertion.ViewAssertions.matches;
-import static android.support.test.espresso.matcher.ViewMatchers.hasSibling;
-import static android.support.test.espresso.matcher.ViewMatchers.isAssignableFrom;
-import static android.support.test.espresso.matcher.ViewMatchers.isChecked;
-import static android.support.test.espresso.matcher.ViewMatchers.isDescendantOfA;
-import static android.support.test.espresso.matcher.ViewMatchers.isDisplayed;
-import static android.support.test.espresso.matcher.ViewMatchers.withContentDescription;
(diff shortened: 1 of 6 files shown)


passsy:
What if loading the task (could theoretically happen from the database or the network) doesn't finish before the Activity receives the first configuration change? The current implementation wouldn't populate the task again because the fragment is not null although the task isn't loaded. 

The `shouldLoadDataFromRepo` should be set when the presenter has finished loading the task. It should also be saved in `onSaveInstanceState()`.

savepopulation:
I think you should focus on repository part to handle orientation change. When a request already made to get tasks you should not send same request just observe the result.

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskActivity.java
```diff
@@ -50,6 +50,9 @@ protected void onCreate(Bundle savedInstanceState) {
         AddEditTaskFragment addEditTaskFragment =
                 (AddEditTaskFragment) getSupportFragmentManager().findFragmentById(R.id.contentFrame);
 
+        // Prevent the presenter from loading data from the repository if this is a config change.
+        boolean shouldLoadDataFromRepo = addEditTaskFragment == null;
```
slinzner: This is very cryptic, i'd rather prefer something like this:
boolean doNotLoadDataOnConfigChange = false;
if {null == aETF} { doNotLoadDataOnConfigChange = true}

This saves you the comment and the reader immediately knows what's going on

todoapp/app/src/androidTest/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksScreenTest.java
```diff
@@ -33,33 +54,11 @@
...
 import static android.support.test.espresso.matcher.ViewMatchers.withId;
 import static android.support.test.espresso.matcher.ViewMatchers.withText;
-
 import static com.example.android.architecture.blueprints.todoapp.TestUtils.getCurrentActivity;
 import static com.google.common.base.Preconditions.checkArgument;
-
 import static org.hamcrest.Matchers.allOf;
 import static org.hamcrest.core.IsNot.not;
 
-import android.support.test.InstrumentationRegistry;
```
slinzner: Was this intentionally or did you just use the wrong formatter?

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskFragment.java
```diff
@@ -86,9 +86,8 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
         View root = inflater.inflate(R.layout.addtask_frag, container, false);
         mTitle = (TextView) root.findViewById(R.id.add_task_title);
         mDescription = (TextView) root.findViewById(R.id.add_task_description);
-
+        setRetainInstance(false);
```
slinzner: false is the default, no need to explicitly set it here

JoseAlcerreca:
@passsy had to read it 5 times to understand what you meant and you're right! We should cover that case in tests. I'll send another commit soon.

passsy:
I'll use simpler sentences in the future 😊

JoseAlcerreca:
My fault, I thought you were talking about something else. PTAL!

passsy:
LGTM 🚢

malmstein:
LGTM!

Issue #229, opened 2016-11-04
All branches: run UI tests on Firebase Test Lab with Travis


JoseAlcerreca:
There's a way to add the client secrets to travis, encrypting a file:

http://engineering.dailymotion.com/running-firebase-test-lab-on-travis-ci/

But I'm not sure about this particular implementation. `after_success` looks like the wrong hook to run UI tests unless we just want to warn about failures. Not sure how that'd work.

oldergod:
As you said, I don't think `after_success` is the right one to run those tests.
I would rather run them as a simple extra step in the `script` list of commands and then use `after_success` or `after_script` to download the results via `gsutil`.

JoseAlcerreca:
Apparently Travis is having problems at the moment. The CLI tool doesn't authenticate
https://github.com/travis-ci/travis.rb/issues/448 and others

PR #230, opened 2016-11-07
Enables running UI tests from CircleCI on Firebase Test Lab

JoseAlcerreca:
For the record: I tried 4 o 5 different ways of decoding the gcloud key. Turns out the base64 string in the env variable must have no spaces. SSHing into the container is neat.

Runs on Nexus10 api 21 and Nexus5X api 25 emulators. 

TODO: 
 - Would devices be faster? 
 - Compatibility check every 24h with real devices from different OEMs?

ping #229

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -18,6 +18,6 @@ dependencies:
 
 test:
   override:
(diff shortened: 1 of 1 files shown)


JoseAlcerreca:
@oldergod can you confirm you signed the SLAs?

oldergod:
Yes I can. I am not sure what I can do with the _CLAs are signed, but unable to verify author consent_ error. I contributed multiple times to GoogleChrome's repository and never saw this error before.

malmstein:
nice one! 🚢

PR #234, opened 2016-11-17
removed StatisticsActivity and replaced the TasksFragment with Statis…

mohamed-habib:
…ticsFragment instead of creating new activity each time i switch between tasks and statistics, now we have TasksActivity and when statistics item is selected the StatisticsFragment is replaced with the TasksFragment and when tasks item is selected the TasksFragment is replaced with the StatisticsFragment.


JoseAlcerreca:
It's one way to do it, but I see no reason for it right now. If there is an issue you want to talk about let's open one. Closing for now.

malmstein:
I agree with @JoseAlcerreca, this is specially set to two `Activity`. Also, take into account the dagger implementation is using `Scopes` per `Activity` and this change would break it.

mohamed-habib:
@googlebot I signed it!

mohamed-habib:
@JoseAlcerreca Thanks for replying, i thought that this is an enhancement to be taken in mind, as i got rid of a boilerplate code that is unnecessary.

mohamed-habib:
@malmstein Thanks for replying, I don't see why it's intended to be two Activities, while the second Activity does exactly the same as the first, i think it'd be better to have one and add the fragment of the second activity to it, isn't this is a reason of why we use fragments in the first place? 
With regard to the dagger, i didn't learn about it yet and i haven't seen it in this branch that's why i didn't put it into account. 
Cheers.

PR #237, opened 2016-11-23
Jack toolchain and lambdas

florina-muntenescu:


todoapp/app/build.gradle
```diff
@@ -12,11 +12,20 @@ android {
...
+
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
     }
 
     buildTypes {
         debug {
-            minifyEnabled true
+            minifyEnabled false
```
JoseAlcerreca: on purpose?

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/TasksRepository.java
```diff
@@ -118,12 +118,7 @@ public static void destroyInstance() {
             // Query the local storage if available. If not, query the network.
             Observable<List<Task>> localTasks = getAndCacheLocalTasks();
             return Observable.concat(localTasks, remoteTasks)
-                    .filter(new Func1<List<Task>, Boolean>() {
-                        @Override
-                        public Boolean call(List<Task> tasks) {
-                            return !tasks.isEmpty();
-                        }
-                    }).first();
+                    .filter(tasks -> !tasks.isEmpty()).first();
```
JoseAlcerreca: nit: move .first to new line.

todoapp/app/build.gradle
florina-muntenescu: Yes. With `minifyEnabled true`, when building, you get the error 
`Minifying the variant used for tests is not supported when using Jack.`

JoseAlcerreca: Uhmm in that case comment and at least minify the release build type. http://stackoverflow.com/questions/37314760/android-how-to-perform-obfuscation-with-the-jack-compiler

It's important to compare APK size and method count.

PR #240, opened 2016-11-25
[todo-mvp-rxjava] added jack compiler and used lambdas

florina-muntenescu:


todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/taskdetail/TaskDetailPresenter.java
```diff
@@ -81,27 +79,14 @@ private void openTask() {
...
-                    }
-
-                    @Override
-                    public void onNext(Task task) {
-                        showTask(task);
-                    }
-                });
-        mSubscriptions.add(subscription);
+                .subscribe(this::showTask,
+                        throwable -> {
```
JoseAlcerreca: It's a super common pattern but since this is educational, I would add a comment per line. // onCompleted // onError // onNext.

vovkab:
Did we do compile time comparison jack vs javac?
Last time I tried jack it was unacceptable slow.

PR #294, opened 2017-03-07
Merges MVP changes into todo-mvp-dagger

JoseAlcerreca:


todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskPresenter.java
```diff
@@ -73,7 +76,7 @@ void setupListeners() {
 
     @Override
     public void start() {
-        if (!isNewTask()) {
+        if (!isNewTask() && mIsDataMissing) {
```
malmstein: I feel like this logic should be in the `Repository` and not in the `Presenter`. How about, instead of passing in data through the `Bundle`, we use `SharedPreferences` to store wether the data needs refresh or not. That way, exposing that information through a `Provider`, the `Repository` can ask the `Provider` if a refresh is needed and trigger it. It can also be tested in the `Repository`, which always makes a puppy happy.

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskPresenter.java
```diff
@@ -112,6 +116,11 @@ public void onDataNotAvailable() {
         }
     }
 
+    @Override
+    public boolean isDataMissing() {
```
malmstein: this is related to above, exposing business logic through the `Presenter` seems like a bad a idea

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskPresenterModule.java
```diff
@@ -31,4 +35,9 @@ public AddEditTaskPresenterModule(AddEditTaskContract.View view, @Nullable Strin
     String provideTaskId() {
         return mTaskId;
     }
+
+    @Provides
+    boolean provideShouldLoadDataFromRepo() {
```
malmstein: why do we need to provide this here?

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskFragment.java
```diff
@@ -86,9 +86,7 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
         View root = inflater.inflate(R.layout.addtask_frag, container, false);
         mTitle = (TextView) root.findViewById(R.id.add_task_title);
         mDescription = (TextView) root.findViewById(R.id.add_task_description);
-
         setHasOptionsMenu(true);
-        setRetainInstance(true);
```
malmstein: and a kitten is now saved from the shelter

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskPresenter.java
JoseAlcerreca: I don't think this is what you think it is. It's to fix a subtle bug that can happen if in a config change the data is not ready yet. It was added to todo-mvp a long time ago.

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskPresenter.java
JoseAlcerreca: It's UI state not business logic.

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskFragment.java
JoseAlcerreca: You can't have it all...

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskPresenter.java
malmstein: why don't we keep it in the `View` then? pass it as a collaborator for example

JoseAlcerreca:
Merged with current todo-mvp to keep this PR fresh.

PR #316, opened 2017-04-03
Todo-mvvm-databinding: "add new task" view only displayed with "All tasks" filter. Fixes #315

### LINKED ISSUE #315: todo-mvvm-databinding: a TextView "Add a TO-DO item +" always visible despite the filter's type
todo-mvvm-databinding

In TasksViewModel I've noticed that you set the current task filtering type, and in switch cases you do tasksAddViewVisible.set(true) set to be true only for the case ALL_TASKS. But that view stays visible in ACTIVE_TASKS, and COMPLETED_TASKS.

JoseAlcerreca:
We were using a @Bindable getter and an Observable field resulting in the same name. (Maybe Data Binding should show a warning if this happens).

Fixes #315.


florina-muntenescu:
LGTM. Feel free to merge it, @JoseAlcerreca

PR #327, opened 2017-04-26
Aligning the setting of the presenter in the constructor, for all views

florina-muntenescu:
Tests added to check that the presenter is set to the view.


JoseAlcerreca:
You're setting the presenter-view association in presenter constructor instead of activity. Some say that you would be violating SRP if the constructor does more than creating the object. I guess that's why the association was being done in the activity (we defined the activity as the controller of the whole presentation layer).

Do you have a better reason to keep it this way?

JoseAlcerreca:
As agreed offline, let's merge this to have consistency and add an issue to fix all of them later.

PR #328, opened 2017-04-26
Moving the setting of the AddEditTaskPresenter in the constructor.

florina-muntenescu:
Tests added to check that the presenter is set to the view.

Fixes [Issue 320](https://github.com/googlesamples/android-architecture/issues/320)


JoseAlcerreca:
As agreed offline, let's merge this to have consistency and add an issue to fix all of them later.

PR #335, opened 2017-05-08
Make DataSource reference in Presenter abstract

EfeBudak:


EfeBudak:
I signed it!

JoseAlcerreca:
We might drop the DataSource from the repository. Usually they don't share interface.

Does this only happen in the tasks presenter?

EfeBudak:
Actually, it happens on every presenter. 
You are right, we can drop the DataSource. However, making it abstract for the presenter is more suitable for dependency inversion. The presenter does not have to know the detail of DataSource.

JoseAlcerreca:
Well let's close this because it should be done in todo-mvp and it's a known issue. Thanks!

savepopulation:
@JoseAlcerreca i think dropping DataSource from repository is not a proper move if you're using it also as a memory cache. It's also a data source because of providing cached data in memory.

JoseAlcerreca:
It's not a source of data. In that case there would have to be a 'cache' DataSource, different from the repo.

Issue #343, opened 2017-06-06
[TODO-Architecture-Component] blueprint


Ajimi:
Can you provide us with Google Android architecture component?

JoseAlcerreca:
dup #337

Issue #337, opened 2017-05-21
[todo-mvvm] Architecture Components sample


gaurav414u:
I want to integrate the latest architectural components (ViewModel, LiveData, LifeCycleOwner) into the todo-mvvm. Should there be a new branch with the name [todo-mvvm-{newandroidarch/somethingelse}] or should I start editing in this branch only?

ber4444:
+1

(also, for reference, he is talking about https://developer.android.com/topic/libraries/architecture/guide.html)

gaurav414u:
Exactly. Thanks!

gaurav414u:
@ber4444 I have started working on it here: https://github.com/gaurav414u/android-architecture/tree/todo-mvvm-databinding_newarch. Please feel free to observe and let me know if something needs to be altered.
Thanks

JoseAlcerreca:
We're working on an official branch for this but feel free to publish yours, we're still defining good and bad patterns and more ideas are welcomed.

gaurav414u:
@JoseAlcerreca , I have started working, but found out that the DataBinding is not lifecycle aware yet. It will be there once  the lifecycle hits 1.0 stable api - as per Yigit.
Should I submit a pull request for this unfinished work? 
I am new over here :)

gaurav414u:
@JoseAlcerreca Can we reopen this issue to discuss on [linkToPullRequestInMyFork](https://github.com/gaurav414u/android-architecture/pull/1)?
Thanks

JoseAlcerreca:
Cool, let's discuss on your fork.

JoseAlcerreca:
We now have dev- branches for both lifecycle-aware components and Room.

https://github.com/googlesamples/android-architecture

PR #346, opened 2017-06-11
[todo-mvvm-databinding_newarch] New branch which uses new android arch components

gaurav414u:
New branch: **todo-mvvm-databinding_newarch**
I have made an attempt to integrate the new arch components into the todo-mvvm-databinding. So, all the ViewModels now extend `android.arch.lifecycle.ViewModel`.
The original plan was to remove the Observables all together and use LiveData only but it came out that Livedata doesnt support databinding as of now. 

- So all the observables which were interacting with data-binding are still as is.
- The snackbarText is implemented using LiveData
- All the @Bindable methods are now removed and are instead implemented as Observables (only for now). Once the android lifecycle hits 1.0 stable api we can remove all the observables and use LiveData only.
- Next, I would want to integrate the `android.arch.persistence.room` for Tasks storage.

### DIFF
--- a/todoapp/app/build.gradle
+++ b/todoapp/app/build.gradle
@@ -57,6 +57,7 @@ android {
     dataBinding {
         enabled = true
     }
+    buildToolsVersion '25.0.0'
 }
 
 /*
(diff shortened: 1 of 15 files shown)


JoseAlcerreca:
Please don't create PR against existing samples, create a new one in your own repo and post it in the issue. Thanks!

gaurav414u:
@JoseAlcerreca Sorry, I didnt get you. You want to me to first create a pull request in my own repo and then post it in the issues #337 and/or #343 ?
Please correct me if I am wrong. Thanks!

PR #353, opened 2017-06-23
Update README.md

goodev:
Link to todo-mvp-rxjava-kotlin


JoseAlcerreca:
There needs to be a review first, can you open an issue instead?

PR #355, opened 2017-06-23
Fix Toast text display position when long press ActionBar menu

### LINKED ISSUE #354: [todo-mvp-rxjava] Menu hint Toast text position is not right
I just look the todo-mvp-rxjava project and find this problem. I think other branches will also have this problem.

toast

goodev:
Fix #354


chrisbanes:
:shipit:

JoseAlcerreca:
Thanks @goodev and @chrisbanes for the review.

I assume this fix needs to be done everywhere, so it should be also applied to todo-mvp. @goodev feel free :)

PR #358, opened 2017-06-25
[dev-todo-mvvm-live] Removed unnecessary parameters from the `saveTask` method 

riggaroo:
The `AddEditTaskViewModel` already contains the title and description of
the task. In order to be a more accurate example of MVVM and the flow of information, the title and description is now
just obtained from the variables in `AddEditTaskViewModel` itself instead of being passed through
from the Fragment.


JoseAlcerreca:
I guess this ended up like this after a refactoring, as it makes no sense :) Thanks!

PR #362, opened 2017-06-27
[todo-mvvm-databinding] Removed unnecessary parameters from the `saveTask` method

riggaroo:
As a more complete example usage of MVVM, I've changed the `saveTask()` method to not take any parameters into the method as the `ViewModel` already has the title and description as class variables. 
Now I am using the values from ViewModel itself instead of passing the values back through to the ViewModel from the Fragment.


JoseAlcerreca:
Thanks again!

PR #376, opened 2017-08-05
Correction in build.gradle

mohitajwani:
Minor referencing issue fixed.

### DIFF
--- a/todoapp/app/build.gradle
+++ b/todoapp/app/build.gradle
@@ -80,7 +80,7 @@ dependencies {
 
     // Android Testing Support Library's runner and rules
     androidTestCompile "com.android.support.test:runner:$rootProject.ext.runnerVersion"
-    androidTestCompile "com.android.support.test:rules:$rootProject.ext.runnerVersion"
+    androidTestCompile "com.android.support.test:rules:$rootProject.ext.rulesVersion"
 
     // Dependencies for Android unit tests
     androidTestCompile "junit:junit:$rootProject.ext.junitVersion"


mohitajwani:
I signed it

Regards,

Mohit Ajwani
(+91-9820217190)

todoapp/app/build.gradle
```diff
@@ -80,7 +80,7 @@ dependencies {
 
     // Android Testing Support Library's runner and rules
     androidTestCompile "com.android.support.test:runner:$rootProject.ext.runnerVersion"
-    androidTestCompile "com.android.support.test:rules:$rootProject.ext.runnerVersion"
+    androidTestCompile "com.android.support.test:rules:$rootProject.ext.rulesVersion"
```
piotrek1543: As both of `rulesVersion` and `runnerVersion` exist in `android-architecture/todoapp/build.gradle` and have the same  `0.5` value (no surprising, they're always have the same) better solution would be to _join them_ into one like `androidTestLibVersion` instead of keeping them both. 

Just a suggestion ;-)

mohitajwani: That's correct. I also feel the same, but I was not sure if these versions can be same always since they were mentioned separately, that is the reason I did not combine them into one.

JoseAlcerreca:
Good catch, thanks!

JoseAlcerreca:
>  better solution would be to join them

Yep, probably! We need to change these things from todo-mvp. I'll add it to the backlog.

mohitajwani:
Although a very small (negligible) contribution, but feeling proud as this is my first commit to the repository. Thanks @JoseAlcerreca

PR #388, opened 2017-08-22
Initial conversion of todo-mvvm-live to kotlin.

DougSig:


todoapp/app/src/androidTest/java/com/example/android/architecture/blueprints/todoapp/ActivityTestExt.kt
```diff
@@ -0,0 +1,80 @@
...
+ */
+// The array is just to wrap the Activity and be able to access it from the Runnable.
+val currentActivity: Activity
+    get() {
+        // The array is just to wrap the Activity and be able to access it from the Runnable.
+        val resumedActivity = arrayOfNulls<Activity>(1)
+        InstrumentationRegistry.getInstrumentation().runOnMainSync {
+            val resumedActivities = ActivityLifecycleMonitorRegistry.getInstance()
+                    .getActivitiesInStage(Stage.RESUMED)
+            if (resumedActivities.iterator().hasNext()) {
```
piotrek1543: what about extracting `resumedActivities.iterator()` to local variable? Actually, you're calling function iterator twice. I don't think it's necessary

todoapp/build.gradle
```diff
@@ -1,10 +1,34 @@
...
+
+    // App dependencies
+    supportLibraryVersion = '26.0.0'
+    guavaVersion = '22.0-android'
+    junitVersion = '4.12'
+    mockitoVersion = '2.7.22'
+    hamcrestVersion = '1.3'
+    runnerVersion = '0.5'
+    rulesVersion = '0.5'
+    espressoVersion = '2.2.2'
```
piotrek1543: there are newer version of android testing library: https://developer.android.com/topic/libraries/testing-support-library/release-notes.html

florina-muntenescu: Increasing the version of the Android Testing Library is not a mandatory requirement in order to use Architecture Components with Kotlin. We prefer to keep the changes to a minimum to ease the branch comparison.

PR #394, opened 2017-08-24
Minor refactors

subhasha1:


todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/taskdetail/TaskDetailActivity.kt
```diff
@@ -33,10 +33,10 @@ class TaskDetailActivity : AppCompatActivity() {
         setContentView(R.layout.taskdetail_act)
 
         // Set up the toolbar.
-        setupActionBar(R.id.toolbar) {
```
DougSig: Keep as original; See https://kotlinlang.org/docs/reference/lambdas.html - "In Kotlin, there is a convention that if the last parameter to a function is a function, and you're passing a lambda expression as the corresponding argument, you can specify it outside of parentheses:"

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/local/TasksLocalDataSource.kt
```diff
@@ -80,22 +81,20 @@ class TasksLocalDataSource private constructor(context: Context) : TasksDataSour
...
                     isCompleted = getInt(getColumnIndexOrThrow(COLUMN_NAME_COMPLETED)) == 1
                 }
-                callback.onTaskLoaded(task)
-            } else {
-                callback.onDataNotAvailable()
             }
             close()
         }
         db.close()
+        task?.let { callback.onTaskLoaded(it) } ?: callback.onDataNotAvailable()
```
DougSig: Hmm, although this is working as intended, I'm not sure this is best practice.
For example:

val test = `expression A`?.let { `expression B` } ?: `expression C`

If `expression A` evaluates to a nonnull and `expression B` evaluates to null, then the value of the left side of the elvis operator is null (because the return value for a `let` block is the last expression inside the block), so `test` would be set to `expression C`. The intention was to set `test` to `expression C` only if `expression A` is null, so that would be incorrect.

Keeping the original would probably be the safest way to go.

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/local/TasksLocalDataSource.kt
```diff
@@ -56,19 +56,20 @@ class TasksLocalDataSource private constructor(context: Context) : TasksDataSour
...
-                callback.onTasksLoaded(tasks)
-            } else {
-                // This will be called if the table is new or just empty.
-                callback.onDataNotAvailable()
-            }
             close()
         }
         db.close()
+
+        if (!tasks.isEmpty()) {
```
DougSig: Bring back to isNotEmpty. Any reason for moving the callbacks to after the cursor.close() and db.close()?

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/local/TasksLocalDataSource.kt
subhasha1: Agreed

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/taskdetail/TaskDetailActivity.kt
subhasha1: Agreed

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/local/TasksLocalDataSource.kt
piotrek1543: @DougSig I think that main reason of this kind of code extracting is that both of callbacks are not strict correlated with database, but with `ArrayList<Task> tasks`. I suppose that an author of commit wanted first copy values from database to tasks, then check if this `List` is empty.

And yes, you're right. Expression` !tasks.isEmpty()` looks a bit weird, restoring previous `tasks.isNotEmpty()` would cause much readable code ;-)

subhasha1: As callback may return at arbitrary time, I thought its better to clean up resources before invoking.

subhasha1: yes @piotrek1543 it should be isNotEmpty, my bad

DougSig: Ok, that's fair.

subhasha1:
revised!

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/local/TasksLocalDataSource.kt
```diff
@@ -94,7 +94,10 @@ class TasksLocalDataSource private constructor(context: Context) : TasksDataSour
             close()
         }
         db.close()
-        task?.let { callback.onTaskLoaded(it) } ?: callback.onDataNotAvailable()
+        if(task!=null){
```
DougSig: Mind correcting the formatting? `if(...){` should be `if (...) {` and `}else` should be `} else`.

SbyXlmm:
大家好

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/local/TasksLocalDataSource.kt
subhasha1: Done

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/local/TasksLocalDataSource.kt
```diff
@@ -80,22 +81,24 @@ class TasksLocalDataSource private constructor(context: Context) : TasksDataSour
...
-                callback.onTaskLoaded(task)
-            } else {
-                callback.onDataNotAvailable()
             }
             close()
         }
         db.close()
+        if(task!=null){
+            task?.let { callback.onTaskLoaded(it) }
+        }else {
```
florina-muntenescu: format the file. Make sure you use the default Android Studio Code Style.

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksFragment.kt
```diff
@@ -270,10 +270,10 @@ class TasksFragment : Fragment(), TasksContract.View {
             with(rowView.findViewById<CheckBox>(R.id.complete)) {
                 // Active/completed task UI
                 isChecked = task.isCompleted
-                rowView.setBackgroundDrawable(viewGroup.context.resources.getDrawable(
+                rowView.setBackgroundResource(
                         if (task.isCompleted) R.drawable.list_completed_touch_feedback
```
florina-muntenescu: extract this into a val for readability

subhasha1: Done!!

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/local/TasksLocalDataSource.kt
subhasha1: Done!! 
Could you please let me know if there is any resource for format guides like newline, tabs and spaces, so that I can add those rules and achieve the format using Reformat Code menu in Android Studio.

subhasha1:
revision 3!!

florina-muntenescu:
@piotrek1543 We’ve noticed that you’ve been reviewing and approving PRs. We strongly support community contributions and review comments that increase the quality of our samples. We do ask you though, not to approve PRs anymore. This is the responsibility of the branch owner(s) and of the maintainers of this repo. By approving PRs we risk confusing the developer that wrote the PR, the owners and the maintainers and as consequence, having PRs that have not been thoroughly reviewed being merged in the sample. Thank you!

piotrek1543:
@florina-muntenescu I understand and I would work on it. Only valuable PR comments, nothing more :-)

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/local/TasksLocalDataSource.kt
```diff
@@ -94,9 +94,11 @@ class TasksLocalDataSource private constructor(context: Context) : TasksDataSour
             close()
         }
         db.close()
-        if(task!=null){
```
DougSig: I had a change of heart about this... I prefer reverting to what you originally had, but use `also` instead of `let` so the left side of `?:` evaluates to the receiver.
Like this: 
task?.also { callbackOnTaskLoaded(it) } ?: callback.onDataNotAvailable()

That way, onDataNotAvailable is always called if `task` is null. The current way technically could silently fail if `task` becomes null between `if (task != null)` and `task?.let ...` (though impossible in this sample).

subhasha1: Done!!
Yes you are right, current code will definitely work for this sample but your suggestion is better as it makes consideration for multi threaded scenarios.

PR #396, opened 2017-08-25
Gradle Version Update

trent-dev:
Gradle Version Update to 2.3.3


abhay222:
You could also update the targetSdkVersion, compileSdkVersion , buildToolsVersion and other dependencies

JoseAlcerreca:
We do the version bumps, thanks.

PR #401, opened 2017-08-29
[Dev todo mvp kotlin] Use more extension functions

GuoTianqi:


todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskActivity.kt
```diff
@@ -51,9 +51,9 @@ class AddEditTaskActivity : AppCompatActivity() {
         var shouldLoadDataFromRepo = true
 
         // Prevent the presenter from loading data from the repository if this is a config change.
-        if (savedInstanceState != null) {
```
DougSig: `if ( localvar != null)` is arguably more readable than `localvar.let` so please revert back to the if statement.

GuoTianqi: I have reverted the commit

DougSig:
Thanks!

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskActivity.kt
arekolek: @DougSig I thought it was being promoted as a [Kotlin idiom](https://kotlinlang.org/docs/reference/idioms.html#execute-if-not-null):

> Execute if not null:
> ```kotlin
> val data = ...
> 
> data?.let {
>     ... // execute this block if not null
> }
> ```

DougSig: ?.<let/apply/also/run> is a language feature, but it comes down to the choice of the programmer and readability of the code. For local vars, using 'if' is safe, and it even makes it easier to have an 'else' clause.

PR #406, opened 2017-08-30
Removes ViewModelItem and small fixes

JoseAlcerreca:
Not sure why the ViewModelItem and related classes were still there, albeit not used.

Lint wasn't complaining because we had unit tests for them :man_facepalming:


JoseAlcerreca:
Thanks! :bowing_man:

PR #415, opened 2017-09-11
Change "Compliment" to "Complement"

mahermalaeb:
Here is why:
[What Is the Difference Between Complement and Compliment?](http://www.write.com/writing-guides/general-writing/word-choice/complement-versus-compliment/)


mahermalaeb:
I signed it!

TwitRco:
There's none, thanks for the fix G

PR #420, opened 2017-09-20
[todo-mvp-rjxava] RxJava 2 migration

kleinsenberg:
Migrate project to RxJava 2.


JoseAlcerreca:
This is why we do the version bumps :) Do you want to fix travis and circle or should I do it?

kleinsenberg:
@JoseAlcerreca fixed 🎉

JoseAlcerreca:
LGTM!  @florina-muntenescu haz los honores

oldergod:
@kleinsenberg I'm trying to understand the implementation of the migration. What's the reasoning for using `Flowable` instead of `Observable` if backpressure is never managed? Is there an meaningful difference between those two?

PR #433, opened 2017-10-13
[todo-mvp-dagger] Removed unnecessary injection call

### LINKED ISSUE #432: [TODO-MVP-DAGGER] ToDoApplication injects itself but doesn't have to
https://github.com/googlesamples/android-architecture/blob/d1cacacded7caf9c572629b0426472fe8758df27/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/ToDoApplication.java#L24-L29

This method should just return the component. The injection is already handled in the parent [`dagger.android.DaggerApplication#onCreate`](https://github.com/google/dagger/blob/master/java/dagger/android/DaggerApplication.java#L53)

oldergod:
closes #432 

`ToDoApplication` does not need to be injected when `applicationInjector()` gets called.

See https://github.com/google/dagger/blob/e8d7cd4c29c1316c5bb1cf0737d4f29111fcb1c8/java/dagger/android/DaggerApplication.java#L56-L61
> Implementations should return an AndroidInjector for the concrete DaggerApplication

The injection is handled in the parent [`dagger.android.DaggerApplication#onCreate`](https://github.com/google/dagger/blob/master/java/dagger/android/DaggerApplication.java#L53)

### DIFF
--- a/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/ToDoApplication.java
+++ b/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/ToDoApplication.java
@@ -23,9 +23,7 @@ public class ToDoApplication extends DaggerApplication {
 
     @Override
     protected AndroidInjector<? extends DaggerApplication> applicationInjector() {
-        AppComponent appComponent = DaggerAppComponent.builder().application(this).build();
-        appComponent.inject(this);
-        return appComponent;
+        return DaggerAppComponent.builder().application(this).build();
     }
 
     /**
--- a/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/ToDoApplication.java
+++ b/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/ToDoApplication.java
@@ -2,15 +2,11 @@ package com.example.android.architecture.blueprints.todoapp;
 
 import android.app.Application;
 import android.support.annotation.VisibleForTesting;
-
 import com.example.android.architecture.blueprints.todoapp.data.source.TasksRepository;
-import com.example.android.architecture.blueprints.todoapp.di.AppComponent;
 import com.example.android.architecture.blueprints.todoapp.di.DaggerAppComponent;
-
-import javax.inject.Inject;
-
 import dagger.android.AndroidInjector;
 import dagger.android.DaggerApplication;
+import javax.inject.Inject;
 
 /**
  * We create a custom {@link Application} class that extends  {@link DaggerApplication}.
--- a/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/di/AppComponent.java
+++ b/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/di/AppComponent.java
@@ -1,18 +1,14 @@
 package com.example.android.architecture.blueprints.todoapp.di;
 
 import android.app.Application;
-
 import com.example.android.architecture.blueprints.todoapp.ToDoApplication;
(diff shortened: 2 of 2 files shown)


digitalbuddha:
@oldergod thank you for contribution but this won't quite work.  Please run ./gradlew connectedAndroidTest locally and you'll see that there are a bunch of NPE coming from `TaskScreenTest#BeforeActivityLaunch` let me know if you want to fix how the `TaskRepository` is loaded during connected tests or if you'd like me to take a look

digitalbuddha:
One thing that you can do is retain the instance of an `AppComponent` within `TodoApplication` and then get the `TasksRepository` from it during testing.  

```java
public class ToDoApplication extends DaggerApplication {

    private AppComponent appComponent = DaggerAppComponent.builder().application(this).build();
    /**
     * Our Espresso tests need to be able to get an instance of the {@link TasksRepository}
     * so that we can delete all tasks before running each test, we do this by creating
     * a single app component instance and allowing us to do
     * taskRepository = application.applicationInjector().taskRepository() from a test
     */

    @Override
    @VisibleForTesting
    public AppComponent applicationInjector() {
        return appComponent;
    }
}

public class TasksScreenTest {

    @Rule
    public ActivityTestRule<TasksActivity> mTasksActivityTestRule =
            new ActivityTestRule<TasksActivity>(TasksActivity.class) {
                @Override
                protected void beforeActivityLaunched() {
                    super.beforeActivityLaunched();
                    // Doing this in @Before generates a race condition.
                    ToDoApplication toDoApplication = (ToDoApplication) InstrumentationRegistry.getTargetContext()
                            .getApplicationContext();
                    toDoApplication.applicationInjector().tasksRepository();
                }
            };
```

digitalbuddha:
One thing that you can do is retain the instance of an `AppComponent` within `TodoApplication` and then get the `TasksRepository` from it during testing.  

```java
public class ToDoApplication extends DaggerApplication {

    private AppComponent appComponent = DaggerAppComponent.builder().application(this).build();
    /**
     * Our Espresso tests need to be able to get an instance of the {@link TasksRepository}
     * so that we can delete all tasks before running each test, we do this by creating
     * a single app component instance and allowing us to do
     * taskRepository = application.applicationInjector().taskRepository() from a test
     */

    @Override
    @VisibleForTesting
    public AppComponent applicationInjector() {
        return appComponent;
    }
}

public class TasksScreenTest {

    @Rule
    public ActivityTestRule<TasksActivity> mTasksActivityTestRule =
            new ActivityTestRule<TasksActivity>(TasksActivity.class) {
                @Override
                protected void beforeActivityLaunched() {
                    super.beforeActivityLaunched();
                    // Doing this in @Before generates a race condition.
                    ToDoApplication toDoApplication = (ToDoApplication) InstrumentationRegistry.getTargetContext()
                            .getApplicationContext();
                    toDoApplication.applicationInjector().tasksRepository();
                }
            };
```

JoseAlcerreca:
I know we asked to keep the UI tests the same between branches but if you want to change how the tests are configured, I think it's a selling point for dagger so feel free to propose things.

By the way, I'm updating some things in https://github.com/googlesamples/android-architecture/pull/455

oldergod:
I'm sorry, I should have run the tests before submitting the PR.

There is no need to change the configuration of the tests.
The type argument of the `AppComponent` was set to `DaggerApplication`. This means while `TodoApplication` is supposed to inject its members, it only does it so far as its `DaggerApplication` scope (if it makes sense).

If the type argument is set to `TodoApplication`, the application would, as expected, inject its members on `onCreate()`, including the `tasksRepository` used in tests.

The fix does this and also removes unnecessary methods in `AppComponent`.

PR #451, opened 2017-11-07
Added syntax format to first snippet inside README

pandres95:


pandres95:
@googlebot I signed it.

PR #472, opened 2017-12-03
Changing access modifiers in UseCaseThreadPoolScheduler to private

shaunpanjabi:
These access modifiers can be private. Not sure why these would need to be public.


shaunpanjabi:
@googlebot I signed it!

PR #475, opened 2017-12-11
A fix for the memory leak issue

tylersuehr7:
The main purpose of this pull request is it to prevent leaking memory whenever we execute use cases.

A memory leak will occur if we execute a long-running use case, that references the presenter's view in the callback of the use case (anonymous inner-class), and the implementation of the view (Activity/Fragment/View)'s lifecycle has ended before the completion of the use case.

The best way that I have found to prevent the above mentioned memory leak is by manually removing the reference to the presenter's view in the appropriate Android framework lifecycle callback for each implementation of the view.

### DIFF
--- a/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/AbstractBasePresenter.java
+++ b/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/AbstractBasePresenter.java
@@ -1,0 +1,107 @@
+package com.example.android.architecture.blueprints.todoapp;
+
+/**
+ * This is the parent of all presenters in this project; it also provides the most
+ * fundamental implementation of {@link BasePresenter}. The main purpose of this class
+ * is it to prevent leaking memory whenever we execute use cases.
+ *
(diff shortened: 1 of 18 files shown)


JoseAlcerreca:
Hey tylersuehr7

Thanks for working on this (next time maybe ping me before embarking on such a big change :smile: )

1. You've made many changes unrelated to the branch. Renaming views in presenters to mView for example. Please read the contribution guidelines: changes need to be kept to a minimum to facilitate comparison between branches and maintenance. Even if you bring improvements, they should be done in the parent branch and we limit those to keep maintenance load low. 

2. The AbstractBasePresenters has a javadoc where you use the first person and add an author. That doesn't age well :) we give credit to contributors in a different way.

> The best way that I have found to prevent the above...

A convenient way to prevent the memory leak is...

3. IView is very .NETy!

First, let's work on reducing the diff size and we'll start a proper review.

Thanks!

tylersuehr7:
Hey Jose,

Thanks for your response and feedback!! I'll heed your advice and update based on your feedback! Thanks!

PR #485, opened 2017-12-28
Endless refreshing on task detail page

eugenebrusov:
Here's how it worked before:

Here's how it works now:


JoseAlcerreca:
Seems like the detail layout is observing this field and the change makes sense. We need to check the Java sample.
However, the field should be a val, not a var.
@DougSig please take a look and feel free to merge.

Thanks!

eugenebrusov:
Okay agree, var changed to val for dataLoading property

PR #486, opened 2017-12-29
less one @Inject

xuie0000:
less one @Inject


JoseAlcerreca:
Please add more details and revert the change to the build.gradle file.

PR #517, opened 2018-03-05
Removes redundant listeners between a Fragment and an Activity

igorgur:
Seems like the code includes callbacks `(*Navigator.java)` which act as a middle man between the Fragments and Activities. That approach should be fine for other architectures and for the 'old' way that has been used. ViewModel should dispatch events to observers and that exactly implemented in the sample code, but the unused interfaces cause a confusion in readability.


igorgur:
Resolved...

igorgur:
Resolved

igorgur:
I signed it!

PR #523, opened 2018-03-06
add cache class to separate caching logic from TasksRepository Fixes #522

### LINKED ISSUE #522: [todo-mvp] Add TasksCacheDataSource
I've noticed that the cache and the caching logic is in the TasksRepository class which, I believe, violates the single responsibility principle. Would it be worth it to create an abstraction for the cache by adding another class? It wouldn't follow the TasksDataSource interface as its functionality is a little different but it would remove a responsibility from TasksRepository and get rid of a lot of if statements that are use for null checks.

I'm going to implement a version of this and see if it works.

jeffmcnd:
This commit adds a new class that handles Task caching. This change
pulls the logic for caching out of the TasksRepository to more closely
adhere to the Single Responsibility Principle (SRP). Now the repository
need not worry about null cache objects, iterating through the cache, or
keeping track of whether the cache is dirty or not.


jeffmcnd:
I signed it!

PR #529, opened 2018-03-20
Link was dead. So was deleted

senpl:
I also think this article replace old one that was referencing.


JoseAlcerreca:
The correct link is https://github.com/googlesamples/android-architecture/tree/deprecated-todo-databinding

senpl:
Main site(https://github.com/googlesamples/android-architecture) stated that "todo-databinding" is Replaced by todo‑mvvm‑databinding . Then in future it will be self reference. That way it seems like link to be deleted.

PR #535, opened 2018-03-29
Fix nullability for Task model

### LINKED ISSUE #512: [dev-todo-mvvm-live-kotlin] App crashes when adding task with empty title or description
`saveTask()` function of `AddEditTaskViewModel.kt` is calling `Task(title.get(), description.get())`  which causes crash when adding new task with title or description being empty.

mailiam:
makes `title` and `description` optional and check emptiness with `CharSequence?.isNullOrBlank()`

fixes #512


mailiam:
I signed it!

hendra4ditya:
#701

emili1996:
Hi  @mailiam would not be better to put the null check when saving the task and never store null titles or descriptions in the database? Something like that.
```
val nonNullTitle = title.get().orEmpty()
val nonNullDescription = description.get().orEmpty()

 val task = Task(nonNullTitle, nonNullDescription)
        if (task.isEmpty) {
            showSnackbarMessage(R.string.empty_task_message)
            return
        }
```

mailiam:
Dear @eisamat,
Thank you for your suggestion. I had to rethink from the scratch since it has been a while since I have first made the contribution. 
If I'm not mistaken current implementation doesn't store empty task to the database(repository).
It only create a Task class with nullable properties in the memory to check emptiness. 
I think its more universal to have null check in the model class. 
Please elaborate more if you think otherwise, I might have misunderstood your code.

PR #544, opened 2018-04-17
Fixed dev-todo-mvvm-rxjava: tasks are never saved to local data source when refreshTasks() is called in TasksRepository

### LINKED ISSUE #543: dev-todo-mvvm-rxjava: tasks are never saved to local data source when refreshTasks() is called in TasksRepository
I followed the example TasksRepository from `dev-todo-mvvm-rxjava` branch, specifically the refresh method where the response from the server is persisted to the local data source and found out an issue with the code. 

`TasksRepository.refreshTasks()` is chained with `.doOnNext(mTasksLocalDataSource::saveTasks)`. Tasks are never saved to local data source because the observable (Completable in this case) that is returned from  `TasksLocalDataSource.saveTasks(tasks)` is never subscribed.

The issue is that `doOnNext()` doesn't subscribe the observable and should only be used for side effects. Instead you should use `flatMap` or similar operator so that the observable will be subscribed as part of the stream.

Nimrodda:
Fixed #543: dev-todo-mvvm-rxjava: tasks are never saved to local data source when refreshTasks() is called in TasksRepository.

I had to update RxJava version so that we now have `flatMapCompletable`.
There were 2 more places in the code that had the same error in addition to the one mentioned in the issue I reported. I fixed those too.

### DIFF
--- a/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/TasksRepository.java
+++ b/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/TasksRepository.java
@@ -169,7 +169,7 @@ public class TasksRepository implements TasksDataSource {
     public Completable refreshTasks() {
         return mTasksRemoteDataSource.getTasks()
                 .subscribeOn(mBaseSchedulerProvider.io())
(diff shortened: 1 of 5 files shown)


Nimrodda:
I signed it!

PR #563, opened 2018-06-23
mvvm-live-kotlin: updated versions: gradle, build tools, several other depends, and res…

alanwgeorge:
…olved issues (mostly around nullability)


alanwgeorge:
I want to move this to it's on branch...

PR #582, opened 2018-08-23
update gradle

MohamedISoliman:
- update  gradle version
- use implementation keyword 
- update build version to be consistent with the gradle plugin version


JoseAlcerreca:
We'll update version periodically, thanks

PR #585, opened 2018-09-03
Fixed wrong methods

yasincidem:
1)Objects class doesn't have the method 'equal' instead it has the method 'equals'.
2)hashCode method is used for a single object. Un this particular code, because we have a sequence of sequence of objects we need to use Objects.hash() method.


yasincidem:
I signed it!

PR #605, opened 2018-10-25
Fix "Failed to Resolve: runtime"

nimdokai:
As mentioned in: https://stackoverflow.com/a/50758769/5864033

>yes ,
If you are getting error like error run time you can change the position of google() in dependencies in build. gradle.. Like below:

`repositories { 
google()
 jcenter()
}`


nimdokai:
I signed it!

Issue #562, opened 2018-06-20
Configuration 'androidTestCompile' is obsolete 


bmomani1:
I got this error When I run todoapp
Configuration 'androidTestCompile' is obsolete and has been replaced with 'androidTestImplementation' and 'androidTestApi'. It will be removed at the end of 2018. For more information see: http://d.android.com/r/tools/update-dependency-configurations.html

florina-muntenescu:
Some of the branches have already been updated by the community. Feel free to create PRs that update the build.gradle file for other branches.

CoolMind:
@Emyboy, see https://makecodesimpleandeasy.blogspot.com/2018/07/configuration-androidtestcompile-is.html.

In `app/build.gradle` do following steps, change:

1) `compile` to `implementation`,
2) `testCompile` to `testImplementation`,
3) `androidTestCompile` to `androidTestImplementation`.

ryanzhou7:
@Emyboy Please see my pull request https://github.com/googlesamples/android-architecture/pull/619
(todo-mvp-dagger branch)

PR #609, opened 2018-11-01
Update to be deprecated in '18 gradle configs from compile to implementation

ryanzhou7:
Android Studio 3.2.1 warns compile configurations will be deprecated in Gradle in 2018. 
See https://developer.android.com/studio/build/dependencies


ryanzhou7:
Closes #562 on todo-mvp branch

PR #614, opened 2018-11-21
[Issue#606] Address build sync issue

jmuci:
Build fails to sync as reported here https://github.com/googlesamples/android-architecture/issues/606


JoseAlcerreca:
Already done, thanks

PR #619, opened 2018-12-14
Upgrade gradle, app dependency versions, warnings for dependency conf…

ryanzhou7:
…igurations (compile to implementation, etc). Addresses #562 on todo-mvp-dagger branch

### DIFF
--- a/todoapp/app/build.gradle
+++ b/todoapp/app/build.gradle
@@ -2,7 +2,7 @@ apply plugin: 'com.android.application'
 
 android {
     compileSdkVersion rootProject.ext.compileSdkVersion
-    buildToolsVersion rootProject.ext.buildToolsVersion
+    buildToolsVersion '28.0.3'
 
     defaultConfig {
         applicationId "com.example.android.architecture.blueprints.todomvpdagger"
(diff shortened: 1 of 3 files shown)


hendercine:
Why hasn't this PR been accepted?

todoapp/app/build.gradle
```diff
@@ -2,7 +2,7 @@ apply plugin: 'com.android.application'
 
 android {
     compileSdkVersion rootProject.ext.compileSdkVersion
-    buildToolsVersion rootProject.ext.buildToolsVersion
+    buildToolsVersion '28.0.3'
```
NalediMadlopha: We have bumped up the build tools version, so there is no need to hardcode the version here. May I suggest we use the `buildToolsVersion` variable so that we only need to update the version in one place when the need arises.

```suggestion
    buildToolsVersion rootProject.ext.buildToolsVersion
```

NalediMadlopha:
> @googlebot I am ok with my commit being contributed to the project. I committed @NalediMadlopha suggestions, I believe he might need to confirm too.

Hey @ryanzhou7, the build is failing. The changes being made is only in the `build.gradle` files as well as the `gradle.properties` file. The change that I suggested in in the `.travis.yml` file. 

I do not mind pushing the changes onto this PR unless you want to do it. Should go ahead or you'll do it?

ryanzhou7:
@NalediMadlopha If you could do it that would be great! I am still new to this would appreciate your expertise.

NalediMadlopha:
@ryanzhou7, I have open a pull request #621 to fix the issue, please check it out.

ryanzhou7:
@NalediMadlopha thanks! So the correct thing for me to do is to close this PR, correct?

JoseAlcerreca:
We bump versions following a process that require merging branches. Please don't send PRs to update this. Thanks!

NalediMadlopha:
@JoseAlcerreca, I'm a bit lost with regards to the context of your comment. May you please clarify what you mean when you say "Please don't send PRs to update this."

Do you mean we should not update the versions of the dependencies or do you mean we should not update the version of the SDK in the `.travis.yml` file?

dazza5000:
I think he is referring to the travis.yml file. Perhaps all of the projects need to be updated at the same time?

PR #623, opened 2019-01-23
Switched the order of dependencies to fix missing manifest merger

weeeBox:
More info: https://stackoverflow.com/questions/50563783/could-not-find-manifest-merger-jar-com-android-tools-buildmanifest-merger26-1


JoseAlcerreca:
It's been done already, thanks!

PR #625, opened 2019-01-24
Changed the order of repositories lists

weeeBox:
Fixes the following error:
`ERROR: Could not find manifest-merger.jar (com.android.tools.build:manifest-merger:26.0.0).`


hqglichao:
thank you

PR #631, opened 2019-02-07
Fix #574 - config changes bug with localized resources - No more AndroidViewModels

### LINKED ISSUE #574: dev-todo-mvvm-live should not be resolving strings in ViewModel
ViewModel should not be resolving strings from resources as shown in 
https://github.com/googlesamples/android-architecture/blob/dev-todo-mvvm-live/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/statistics/StatisticsViewModel.java

If user changes locale on device, activity/fragment will be recreated but ViewModel will be retained with data, that might be invalid for new configuration/locale. Because this is sample referenced by official google documentation, it should not be showcasing invalid ViewModel usage.

ref. https://issuetracker.google.com/u/1/issues/111961971

JoseAlcerreca:
Exposing resolved resources from ViewModels is not a good idea because they survive configuration changes, including locale changes.  See #574 and https://issuetracker.google.com/issues/111961971

In this PR we leverage Data Binding to load those resources for us with the nice side-effect of removing the need for AndroidViewModels so no context is available in them. (I'm considering moving AndroidViewModels to the smell category.)

### DIFF
--- a/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksViewModel.java
+++ b/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksViewModel.java
@@ -16,10 +16,6 @@
 
 package com.example.android.architecture.blueprints.todoapp.tasks;
 
-import android.app.Application;
-import android.content.Context;
-import android.graphics.drawable.Drawable;
-
 import com.example.android.architecture.blueprints.todoapp.Event;
 import com.example.android.architecture.blueprints.todoapp.R;
 import com.example.android.architecture.blueprints.todoapp.addedittask.AddEditTaskActivity;
@@ -34,10 +30,10 @@ import java.util.List;
 import androidx.arch.core.util.Function;
 import androidx.databinding.BaseObservable;
 import androidx.databinding.Bindable;
-import androidx.lifecycle.AndroidViewModel;
 import androidx.lifecycle.LiveData;
 import androidx.lifecycle.MutableLiveData;
 import androidx.lifecycle.Transformations;
+import androidx.lifecycle.ViewModel;
 
 
 /**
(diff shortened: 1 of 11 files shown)


todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksViewModel.java
```diff
@@ -47,17 +43,17 @@
...
     private final MutableLiveData<Boolean> mDataLoading = new MutableLiveData<>();
 
-    private final MutableLiveData<String> mCurrentFilteringLabel = new MutableLiveData<>();
+    private final MutableLiveData<Integer> mCurrentFilteringLabel = new MutableLiveData<>();
 
-    private final MutableLiveData<String> mNoTasksLabel = new MutableLiveData<>();
+    private final MutableLiveData<Integer> mNoTasksLabel = new MutableLiveData<>();
 
-    private final MutableLiveData<Drawable> mNoTaskIconRes = new MutableLiveData<>();
+    private final MutableLiveData<Integer> mNoTaskIconRes = new MutableLiveData<>();
```
Milack27: @JoseAlcerreca Is it really needed to expose these objects as _Mutable_ LiveData?

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksViewModel.java
```diff
@@ -169,15 +156,15 @@ public void completeTask(Task task, boolean completed) {
...
 
-    public LiveData<String> getNoTasksLabel() {
+    public MutableLiveData<Integer> getNoTasksLabel() {
         return mNoTasksLabel;
     }
 
-    public LiveData<Drawable> getNoTaskIconRes() {
+    public MutableLiveData<Integer> getNoTaskIconRes() {
         return mNoTaskIconRes;
     }
```
Milack27: @JoseAlcerreca Is it really needed to expose these objects as _Mutable_ LiveData?

JoseAlcerreca: No, they should be MutableLiveDatas only in the ViewModel, exposed as LiveDatas. That's a bug.

PR #636, opened 2019-03-02
Update TaskViewModel.java

pnurep:
Reduced the number of calls :)


pnurep:
I signed it!

todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/TaskViewModel.java
```diff
@@ -116,10 +116,12 @@ public boolean isDataLoading() {
     // This could be an observable, but we save a call to Task.getTitleForList() if not needed.
     @Bindable
     public String getTitleForList() {
-        if (mTaskObservable.get() == null) {
+        Task task = mTaskObservable.get();
+
+        if (task == null)
```
JoseAlcerreca: We don't do multiline ifs without braces!

PR #651, opened 2019-04-03
Use a consistent process for attaching Fragment to Activity

agodding:
This addresses two issues

1. Use a consistent process across the app's four Activities for attaching the primary Fragment. I've used the existing naming and style from TasksActivity elsewhere. The Fragment is now always configured in a function called `setupViewFragment()`.
2. Only invoke `replaceFragmentInActivity()` if the Fragment is not already present in the FragmentManager. This prevents re-attaching the same Fragment that the system has already recreated upon configuration changes.


agodding:
I signed it!

PR #676, opened 2019-10-01
Omit verbose return type & simplify return syntax

NahroTo:
Return type is already clear by the method name.


NahroTo:
@googlebot I signed it!

PR #680, opened 2019-10-25
Fix DefaultTasksRepository source preference

dhruvnagarajan:
- Give local source preference over remote source while fetching Tasks
If force update is not requested, repo should try to fetch the Task from local source first.

- Refactor and separate repository 'outbound request' (interface) from implementation
Interfaces and Implementations should be in their own packages under domain/repository, and app/data/repository.
If we setup a domain module, then Repository Interfaces under domain/repository could be moved to ":domain" along with domain/usecases


dhruvnagarajan:
@googlebot I signed it!

PR #683, opened 2019-11-12
Upgraded the kotlin version and some gradle dependency version ( exce…

recker115:
…pt the testing ones )


recker115:
@googlebot I signed it!

PR #686, opened 2019-11-17
add stores to Tasks View Model TODO fix Task View Model

digitalbuddha:


digitalbuddha:
oops sorry opened against wrong repo :-)

PR #694, opened 2019-11-29
Migrates the usecases branch to a reactive arch 

JoseAlcerreca:


roel1975:
Sorry for waiting. How do you check this? I think I know. Hang on

PR #732, opened 2020-04-28
Remove unused singleton imports

sssurvey:
Remove the import:
```kotlin
import javax.inject.Singleton
```
I believe the annotation was there for the ```ViewModelFactory.kt ``` and then removed due to similar issue in this link: https://github.com/google/dagger/issues/1607


piotrek1543:
👍 +1

PR #748, opened 2020-07-18
hide keyboard when press navigation up button

michaelsam94:


michaelsam94:
@googlebot I signed it

PR #755, opened 2020-08-06
Added viewLifecycleOwner to the observers inside Fragments

gastsail:
Added viewLifecycleOwner to the observers inside fragments to handle the viewmodel operations withing the lifecycle of the Fragment


gastsail:
Can we check on this @yigit ? Thanks

PR #791, opened 2021-06-01
Migrates CI to GHA

JoseAlcerreca:


JoseAlcerreca:
this is in preparation for the update to 4.2, so merging for now so we can test in the PR

PR #792, opened 2021-06-01
Merge main into dev-hilt

JoseAlcerreca:


.github/workflows/blueprints.yaml
```diff
@@ -0,0 +1,84 @@
+name: blueprints
+
+on:
+  push:
+    branches:
+      - dev-hilt
+      - hilt
```
manuelvicnt, line 7: Nit: we just need `dev-hilt` here

manuelvicnt (approved):

.github/workflows/blueprints.yaml
JoseAlcerreca: My idea is to rename this branch to `hilt` today and make it official, replacing `android-dagger`

PR #793, opened 2021-06-16
Simplify FilteringPopUpMenu inflation

lucasfukuda:
The inflate() method is equivalent to the previous implementation.


lucasfukuda:
@googlebot I signed it!

PR #799, opened 2021-07-21
Fixed HiltExt

uberchilly:
```HiltExt``` returns ```ActivityScenario``` so that tests can use ```moveToState``` similar to ```launchInContainer```


uberchilly:
@googlebot I signed it!

Issue #507, opened 2018-02-13
"todo-mvp-rxjava" Possible to create empty task


Lebedevsd:
When trying to create a task without name Snackbar appears and save not happening. But if you create a task, and then try to edit, put there an empty name, then the task is being saved with an empty name.

VaradGupta23:
the bug is essentially:

Add flow → validation works: if you try to create a new task with an empty title, the Snackbar warns you and the task isn’t saved.

Edit flow → validation missing: if you open an existing task, delete the title, and save, it does save with an empty title.

Why this happens in todo-mvp-rxjava
In the RxJava branch:

AddEditTaskPresenter (or ViewModel equivalent) has validation logic in the add path:


if (title.isEmpty()) {
    mAddTaskView.showEmptyTaskError();
} else {
    mTasksRepository.saveTask(new Task(title, description));
}
But in the edit path, it’s doing something like:


Task task = mTask; // existing
task.setTitle(title);
task.setDescription(description);
mTasksRepository.saveTask(task);
…without checking if the new title is empty.

How to fix it
We just need the same validation for edits as for adds.

Before

private void saveTask(String title, String description) {
    if (mIsNewTask) {
        createTask(title, description);
    } else {
        updateTask(title, description);
    }
}
After

private void saveTask(String title, String description) {
    if (title == null || title.trim().isEmpty()) {
        mAddTaskView.showEmptyTaskError();
        return;
    }

    if (mIsNewTask) {
        createTask(title, description);
    } else {
        updateTask(title, description);
    }
}
This way both create and update paths are blocked if the title is empty.

Why it’s important
Keeps data integrity consistent — you can’t bypass validation just by editing.

Matches the user experience between Add and Edit flows.

Prevents persisting invalid tasks that might cause issues elsewhere in the app.

PR #804, opened 2021-08-20
Resolved issue #507

Samyakjain003:
When trying to create a task empty name or empty title, Snackbar appears and save not happening.And if you create a task, and then try to edit, then the task will not be able to save with an empty name or empty title and snackbar appears and says "Tasks cannot be empty".
#507 Resolved issue "todo-mvp-rxjava" Possible to create empty task.


app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskViewModel.kt
```diff
@@ -104,6 +104,10 @@ class AddEditTaskViewModel(
         val currentTitle = title.value
         val currentDescription = description.value
 
+        if(currentTitle.isNullOrBlank()|| currentDescription.isNullOrBlank()){
```
parmeshtoyou, line 107: looks like code formatting is missing in this line.

PR #806, opened 2021-09-17
Update all library and plugin versions

mohitajwani:


mohitajwani:
@JoseAlcerreca Merge requested

PR #821, opened 2022-03-03
Migrate Statistics to Compose

manuelvicnt:


app/src/sharedTest/java/com/example/android/architecture/blueprints/todoapp/statistics/StatisticsScreenTest.kt
```diff
@@ -0,0 +1,73 @@
...
+@ExperimentalCoroutinesApi
+class StatisticsScreenTest {
+
+    @get:Rule
+    val composeTestRule = createAndroidComposeRule<ComponentActivity>()
+
+    @Test
+    fun tasks_showsNonEmptyMessage() {
+        // Given some tasks
+        val repository = FakeRepository().apply {
```
JoseAlcerreca, line 51: I'm wondering what our recommendation should be here. You can make a case for a unit test that calls the stateless StatisticsContent. It has less coverage but it's much more focused.

JoseAlcerreca (approved):

PR #822, opened 2022-03-03
Migrate TaskDetail to Compose

manuelvicnt:


app/src/androidTestMock/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksActivityTest.kt
```diff
@@ -63,6 +70,10 @@ class TasksActivityTest {
 
     private lateinit var repository: TasksRepository
 
+    @get:Rule
+    val composeTestRule = createAndroidComposeRule<TasksActivity>()
+    private val activity by lazy { composeTestRule.activity }
```
bentrengrove, line 75: I guess it doesn't really matter but wouldn't `get()=` be technically more correct? Assuming you are using this just to delay the reference to activity

app/src/main/java/com/example/android/architecture/blueprints/todoapp/taskdetail/TaskDetailFragment.kt
```diff
@@ -22,86 +22,71 @@ import android.view.MenuInflater
...
+                }
+            }
+        }
     }
 
     override fun onOptionsItemSelected(item: MenuItem): Boolean {
         return when (item.itemId) {
             R.id.menu_delete -> {
-                viewModel.deleteTask()
+                lifecycleScope.launch {
```
bentrengrove, line 84: This new scoping here feels a bit weird to me but maybe it's ok because I am not a coroutine expert. Wouldn't this now cause a delay while it waits to delete the task? I'm sure it's super quick but given this is an architecture sample would it be better to architect this as if it was slow like a network call? 

The other thing that feels wrong is that navigate call would have the effect of cancelling the scope of this coroutine, is that fine?

app/src/test/java/com/example/android/architecture/blueprints/todoapp/taskdetail/TaskDetailViewModelTest.kt
```diff
@@ -139,17 +140,7 @@ class TaskDetailViewModelTest {
     }
 
     @Test
-    fun clickOnEditTask_SetsEvent() {
```
bentrengrove, line 142: Just checking this test isn't needed anymore? I guess it goes with the edit task todo above?

app/src/androidTestMock/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksActivityTest.kt
manuelvicnt: I guess both could work, I don't have a strong preference there, I'm just keeping it consistent with the other Compose samples we have out there

app/src/main/java/com/example/android/architecture/blueprints/todoapp/taskdetail/TaskDetailFragment.kt
manuelvicnt: Yeah, that's fine. By the time the navigation call happens, the task is already deleted since `deleteTask` is a suspend function.

> Wouldn't this now cause a delay while it waits to delete the task?
Agreed, but I'm not modifying ViewModel business logic in this migration, but if the `deleteTask` call could take a long time, the ViewModel should update the `uiState` to indicate that something is in progress. Maybe that's something to fix in the future as the same thing could've happened before.

app/src/test/java/com/example/android/architecture/blueprints/todoapp/taskdetail/TaskDetailViewModelTest.kt
manuelvicnt: Not needed, as this is a UI test for the screen, we shouldn't check for their transitions. That's something we're covering in the end2end tests of AppNavigation and Activity tests

app/src/main/java/com/example/android/architecture/blueprints/todoapp/taskdetail/TaskDetailScreen.kt
```diff
@@ -0,0 +1,230 @@
...
+        // similar to AddEditTaskState
+
+        // Start loading data
+        LaunchedEffect(Unit) {
+            viewModel.start(taskId)
+        }
+
+        // Process snackbar events
+        val snackbarEvent by viewModel.snackbarText.observeAsState()
+        snackbarEvent?.getContentIfNotHandled()?.let { messageId ->
```
JoseAlcerreca, line 92: Why is this not migrated to current guidance? A shared event? for migrations like these I'd like to see TODOs

app/src/main/java/com/example/android/architecture/blueprints/todoapp/taskdetail/TaskDetailScreen.kt
```diff
@@ -0,0 +1,230 @@
...
+                Task("Title", "Description"),
+                taskCompleted = false,
+                onTaskCheck = { },
+                onRefresh = { }
+            )
+        }
+    }
+}
+
+//@Preview
```
JoseAlcerreca, line 222: comment or remove

app/src/main/java/com/example/android/architecture/blueprints/todoapp/taskdetail/TaskDetailFragment.kt
bentrengrove: All good!

app/src/androidTestMock/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksActivityTest.kt
manuelvicnt: Sure! Done

app/src/androidTestMock/java/com/example/android/architecture/blueprints/todoapp/tasks/AppNavigationTest.kt
```diff
@@ -66,6 +68,7 @@ class AppNavigationTest {
 
     @get:Rule
     val composeTestRule = createAndroidComposeRule<TasksActivity>()
+    private val activity by lazy { composeTestRule.activity }
```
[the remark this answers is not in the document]
manuelvicnt: Done! Thanks

app/src/androidTestMock/java/com/example/android/architecture/blueprints/todoapp/tasks/AppNavigationTest.kt
```diff
@@ -111,6 +114,7 @@ class AppNavigationTest {
 
         // Check that statistics screen was opened.
         composeTestRule.onNodeWithText("You have no tasks.").assertIsDisplayed()
+        composeTestRule.waitForIdle()
```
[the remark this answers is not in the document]
manuelvicnt: This is not needed in the end, I remove all the `waitForIdle` in the last migration PR (that's why I didn't add any comments). I found this make tests more reliable when using data binding and Compose (although maybe it's just a coincidence, not sure)

app/src/androidTestMock/java/com/example/android/architecture/blueprints/todoapp/tasks/AppNavigationTest.kt
```diff
@@ -172,23 +177,25 @@ class AppNavigationTest {
 
     @Test
     fun taskDetailScreen_doubleUIBackButton() {
+        dataBindingIdlingResource.monitorActivity(composeTestRule.activityRule.scenario)
+
         val task = Task("UI <- button", "Description")
```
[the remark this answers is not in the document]
manuelvicnt: What do you mean by extract? This will be gone in the last PR when we remove data binding

app/src/main/java/com/example/android/architecture/blueprints/todoapp/taskdetail/TaskDetailScreen.kt
manuelvicnt: This is migrated in the last PR as well. I'm just migrating UI here, this actually works fine here, it starts to be an issue in the tasks screen.

app/src/androidTestMock/java/com/example/android/architecture/blueprints/todoapp/tasks/AppNavigationTest.kt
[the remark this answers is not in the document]
JoseAlcerreca: I was commenting on the "UI <- button" string

JoseAlcerreca (approved):

PR #823, opened 2022-03-03
Migrate AddEditTasks screen to Compose

manuelvicnt:


app/src/androidTestMock/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksActivityTest.kt
```diff
@@ -104,11 +109,10 @@ class TasksActivityTest {
 
     @Test
     fun editTask() {
-        repository.saveTaskBlocking(Task("TITLE1", "DESCRIPTION"))
+        dataBindingIdlingResource.monitorActivity(composeTestRule.activityRule.scenario)
```
bentrengrove, line 112: What does this do? How come we still need it when you are removing data binding for Compose?

app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskScreen.kt
```diff
@@ -0,0 +1,134 @@
...
+    loading: Boolean,
+    title: String,
+    description: String,
+    onTitleChanged: (String) -> Unit,
+    onDescriptionChanged: (String) -> Unit,
+    modifier: Modifier = Modifier
+) {
+    if (loading) {
+        SwipeRefresh(
+            state = rememberSwipeRefreshState(loading),
```
bentrengrove, line 93: I know loading is always true by this point but I think it might be more clear what you are doing here if you added a comment and also just used `true` instead of `loading` in the refresh state

app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskState.kt
```diff
@@ -0,0 +1,98 @@
...
+    taskId: String?,
+    viewModel: AddEditTaskViewModel,
+    onTaskUpdate: () -> Unit,
+    scaffoldState: ScaffoldState = rememberScaffoldState(),
+    lifecycleOwner: LifecycleOwner = LocalLifecycleOwner.current,
+    context: Context = LocalContext.current,
+    coroutineScope: CoroutineScope = rememberCoroutineScope()
+): AddEditTaskState {
+    val currentOnTaskUpdateState by rememberUpdatedState(onTaskUpdate)
+    return remember(viewModel, scaffoldState, lifecycleOwner) {
```
bentrengrove, line 44: Don't need taskId as a key?

app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskState.kt
```diff
@@ -0,0 +1,98 @@
...
+        viewModel.taskUpdatedEvent.observe(lifecycleOwner) {
+            it.getContentIfNotHandled()?.let {
+                onTaskUpdate()
+            }
+        }
+
+        // Listen for snackbar events
+        viewModel.snackbarText.observe(lifecycleOwner) {
+            it.getContentIfNotHandled()?.let { message ->
+                val snackbarText = context.getString(message)
```
bentrengrove, line 76: Holding a context like this always feels a bit scary but I guess this is ok because the observer is mapped to the lifecycleOwner?

app/src/androidTestMock/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksActivityTest.kt
manuelvicnt: This is an end2end test that still uses the Activity and some other Fragments. We'll be able to remove this in the last PR when everything is migrated to Compose, but data binding is still used in some parts of the app at the moment

app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskScreen.kt
manuelvicnt: Done thanks!

app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskState.kt
manuelvicnt: Yes! And everything else but the lambda :D good catch! I fixed this is another CL that I have locally but forgot to add it here. Done! Thanks

app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskState.kt
manuelvicnt: Yeah, this is ok because `AddEditTaskState` is a state holder scoped to the Composition. Thus, memory leaks can't happen.

bentrengrove (approved):

JoseAlcerreca (approved):

PR #824, opened 2022-03-03
Update test core to 1.4.1-alpha04

alexvanyo:
This allows instrumentation tests (`connected***Test`) to run again.

I did not investigate why Espresso `3.5.0-alpha04` and the previous version of test core resulted in no tests being run.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,7 +44,7 @@ ext {
 
     // App dependencies
     androidXVersion = '1.0.0'
-    androidXTestCoreVersion = '1.4.0'
+    androidXTestCoreVersion = '1.4.1-alpha04'
     androidXTestExtKotlinRunnerVersion = '1.1.3'
     androidXTestRulesVersion = '1.2.0'
     androidXAnnotations = '1.3.0'


manuelvicnt (approved):

build.gradle
```diff
@@ -44,7 +44,7 @@ ext {
 
     // App dependencies
     androidXVersion = '1.0.0'
-    androidXTestCoreVersion = '1.4.0'
+    androidXTestCoreVersion = '1.4.1-alpha04'
     androidXTestExtKotlinRunnerVersion = '1.1.3'
     androidXTestRulesVersion = '1.2.0'
```
TWiStErRob, line 49: This should match Espresso/AndroidX too, @alexvanyo, isn't it?
1.2.0 / 3.2.0, 1.3.0 / 3.3.0, 1.4.0 / 3.4.0, etc.

You have 1.2.0 / 1.4.0 / 3.5.0 for libraries that are coupled.

They come as part of the same release train: https://developer.android.com/jetpack/androidx/releases/test

PR #825, opened 2022-03-07
Update coroutines version to 1.6

zsmb13:
Updates coroutines to 1.6, uses the new test APIs.

Bonus: fixes the UI test that's been broken on CI.

### DIFF
--- a/app/src/sharedTest/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksFragmentTest.kt
+++ b/app/src/sharedTest/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksFragmentTest.kt
@@ -25,9 +25,8 @@ import androidx.recyclerview.widget.RecyclerView
 import androidx.test.core.app.ActivityScenario
 import androidx.test.core.app.ActivityScenario.launch
 import androidx.test.core.app.ApplicationProvider
-import androidx.test.core.app.ApplicationProvider.getApplicationContext
+import androidx.test.espresso.Espresso
 import androidx.test.espresso.Espresso.onView
-import androidx.test.espresso.Espresso.openActionBarOverflowOrOptionsMenu
 import androidx.test.espresso.action.ViewActions.click
 import androidx.test.espresso.assertion.ViewAssertions.doesNotExist
 import androidx.test.espresso.assertion.ViewAssertions.matches
@@ -266,7 +265,7 @@ class TasksFragmentTest {
         launchActivity()
 
         // Click clear completed in menu
-        openActionBarOverflowOrOptionsMenu(getApplicationContext())
+        Espresso.openContextualActionModeOverflowMenu()
         onView(withText(R.string.menu_clear)).perform(click())
 
         onView(withId(R.id.menu_filter)).perform(click())
--- a/app/src/sharedTest/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksFragmentTest.kt
+++ b/app/src/sharedTest/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksFragmentTest.kt
@@ -46,7 +46,6 @@ import com.example.android.architecture.blueprints.todoapp.data.source.TasksRepo
 import com.example.android.architecture.blueprints.todoapp.util.saveTaskBlocking
 import junit.framework.Assert.assertEquals
 import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.test.runBlockingTest
 import org.hamcrest.CoreMatchers.allOf
 import org.hamcrest.Matcher
 import org.hamcrest.core.IsNot.not
@@ -77,7 +76,7 @@ class TasksFragmentTest {
     }
 
     @After
-    fun cleanupDb() = runBlockingTest {
+    fun cleanupDb() {
         ServiceLocator.resetRepository()
     }
 
--- a/app/src/sharedTest/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksFragmentTest.kt
+++ b/app/src/sharedTest/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksFragmentTest.kt
@@ -25,8 +25,8 @@ import androidx.recyclerview.widget.RecyclerView
 import androidx.test.core.app.ActivityScenario
 import androidx.test.core.app.ActivityScenario.launch
 import androidx.test.core.app.ApplicationProvider
-import androidx.test.espresso.Espresso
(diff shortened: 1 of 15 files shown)


app/src/test/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskViewModelTest.kt
```diff
@@ -78,9 +83,9 @@ class AddEditTaskViewModelTest {
     }
 
     @Test
-    fun loadTasks_loading() {
-        // Pause dispatcher so we can verify initial values
-        mainCoroutineRule.pauseDispatcher()
+    fun loadTasks_loading() = runTest {
+        // Override eagerly executing Main dispatcher for just a single test
+        Dispatchers.setMain(StandardTestDispatcher())
```
JoseAlcerreca, line 88: How's this "pausing" the dispatcher?

app/src/test/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksViewModelTest.kt
```diff
@@ -67,35 +71,36 @@ class TasksViewModelTest {
     }
 
     @Test
-    fun loadAllTasksFromRepository_loadingTogglesAndDataLoaded() {
-        // Pause dispatcher so we can verify initial values
-        mainCoroutineRule.pauseDispatcher()
+    fun loadAllTasksFromRepository_loadingTogglesAndDataLoaded() =
```
JoseAlcerreca, line 74: nit: change to brackets as you're indenting anyway

app/src/test/java/com/example/android/architecture/blueprints/todoapp/LiveDataTestUtil.kt
```diff
@@ -63,7 +63,7 @@ fun <T> LiveData<T>.getOrAwaitValue(
 /**
  * Observes a [LiveData] until the `block` is done executing.
  */
-fun <T> LiveData<T>.observeForTesting(block: () -> Unit) {
+suspend fun <T> LiveData<T>.observeForTesting(block: suspend () -> Unit) {
```
JoseAlcerreca, line 66: Revert, LiveData in this case has nothing to do with suspend funs and this class is copypasted a lot.

JoseAlcerreca (changes requested):

app/src/test/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskViewModelTest.kt
manuelvicnt, line 88: Instead of doing this, why not passing the `StandardTestDispatcher` to the `MainCoroutineRule`?

app/src/test/java/com/example/android/architecture/blueprints/todoapp/statistics/StatisticsViewModelTest.kt
```diff
@@ -100,9 +103,9 @@ class StatisticsViewModelTest {
     }
 
     @Test
-    fun loadTasks_loading() {
-        // Pause dispatcher so we can verify initial values
-        mainCoroutineRule.pauseDispatcher()
+    fun loadTasks_loading() = runTest {
+        // Override eagerly executing Main dispatcher for just a single test
+        Dispatchers.setMain(StandardTestDispatcher())
```
manuelvicnt, line 108: Same here, should you pass the `StandardTestDispatcher ` to the `MainCoroutineRule` and then call `advanceUntilIdle` as well in the test method above?

manuelvicnt: If you decide to keep this `setMain` here, maybe improve the comment? explain what `StandardTestDispatcher` does instead of the behavior you're overriding

app/src/test/java/com/example/android/architecture/blueprints/todoapp/LiveDataTestUtil.kt
zsmb13: Good point, done. This turned out to be easier than expected to work around.

app/src/test/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskViewModelTest.kt
zsmb13: All other tests expect eager execution on the Main dispatcher, so some of them would break if we replaced it for the entire test class (or we'd have to make their code more complex by adding things like `advanceUntilIdle`).

The problem here is that with the old APIs we had eager by default which could be cleanly overridden by pause/resume. Now that functionality has be achieved with using a StandardTestDispatcher instead of an UnconfinedTestDispatcher for the main thread in the specific test.

Some alternatives in cases like this would be:
- Using setMain/resetMain directly in each test method, a bit more boilerplate than going with the rule
- Splitting your tests into multiple test classes, depending on which Main dispatcher you need 🙈

zsmb13: Pausing the Main dispatcher is used to stop it from eagerly executing coroutines, and trigger their execution manually instead. This is what StandardTestDispatcher provides in the new APIs (vs UnconfinedTestDispatcher, which is eager).

(See also: [migration guide section about pausing](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-test/MIGRATION.md#replace-usages-of-pausedispatcher-and-resumedispatcher-with-a-standardtestdispatcher))

manuelvicnt (approved):
LGTM! But also wait for @JoseAlcerreca to take a look before merging. Thanks!

app/src/sharedTest/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksFragmentTest.kt
```diff
@@ -267,7 +265,7 @@ class TasksFragmentTest {
         launchActivity()
 
         // Click clear completed in menu
-        openActionBarOverflowOrOptionsMenu(getApplicationContext())
+        Espresso.openContextualActionModeOverflowMenu()
```
johnjohndoe, line 268: This could be a static import.

zsmb13: Done!

JoseAlcerreca (approved):
This is SO much better!

PR #826, opened 2022-03-10
Migrate Tasks to Compose

manuelvicnt:


build.gradle
```diff
@@ -58,7 +58,7 @@ ext {
     composeVersion = '1.1.1'
     coroutinesVersion = '1.5.2'
     dexMakerVersion = '2.12.1'
-    espressoVersion = '3.5.0-alpha04' // This breaks Robolectric which needs Espresso 3.5 alphas
+    espressoVersion = '3.4.0' // This breaks Robolectric which needs Espresso 3.5 alphas
```
JoseAlcerreca, line 61: why the downgrade?

JoseAlcerreca (approved):

build.gradle
manuelvicnt: To be able to run the Espresso tests in the emulator locally, I need to get the upstream changes that fix this. Thanks for the reminder

PR #856, opened 2022-05-26
Experimental functional reactive state production

tunjid:
An experimental branch for functional reactive state production based on the [stateflow branch](https://github.com/android/architecture-samples/tree/stateflow).

It offers feature parity with the existing branch and tries to establish some core FRP state production concepts:

* Idiomatically functional: All inputs are eventually reduced into the UI state
* No side effects: all changes to state must be modeled in the stream
* Restricted UI state visibility: The state production pipeline cannot read the UI state until is ready to make changes to it
* Well defined inputs: All information required to process an action must be present at the input of the action


manuelvicnt:
MVI? Who's MVI? I'm just a functional reactive state production approach!

image

PR #860, opened 2022-06-21
Use new ViewModel factory APIs

manuelvicnt:


JoseAlcerreca (approved):
It's like code from the past!

PR #862, opened 2022-06-25
UDF experiment for producing state by merging flows

tunjid:
Produces state by merging flows that carry functions that lazily modify the state.

This avoids the issues that can arise when combining flows:

* The limited arity of the `combine` function
* The difficulty of merging the effects of user actions into the state production pipeline

There are a probably a few rough edges, but it has an element of consistency across all ViewModels.

This change has the interesting side effect of only being able to push state changes only if there is an observer of UI state. If there is no observer of state, all pending pushes will suspend until an observer is present. This one of the reasons for adding Turbine as a test dependency.

### DIFF
--- a/app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskViewModel.kt
+++ b/app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskViewModel.kt
@@ -24,12 +24,14 @@ import com.example.android.architecture.blueprints.todoapp.TodoDestinationsArgs
 import com.example.android.architecture.blueprints.todoapp.data.Result.Success
 import com.example.android.architecture.blueprints.todoapp.data.Task
 import com.example.android.architecture.blueprints.todoapp.data.source.TasksRepository
+import com.example.android.architecture.blueprints.todoapp.util.Mutation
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskViewModel.kt
```diff
@@ -55,25 +59,23 @@ class AddEditTaskViewModel @Inject constructor(
...
     private val taskId: String? = savedStateHandle[TodoDestinationsArgs.TASK_ID_ARG]
 
-    // A MutableStateFlow needs to be created in this ViewModel. The source of truth of the current
-    // editable Task is the ViewModel, we need to mutate the UI state directly in methods such as
-    // `updateTitle` or `updateDescription`
-    private val _uiState = MutableStateFlow(AddEditTaskUiState())
-    val uiState: StateFlow<AddEditTaskUiState> = _uiState.asStateFlow()
+    private val eventStateChanges = MutableSharedFlow<StateChange<AddEditTaskUiState>>()
 
-    init {
```
tunjid, line 64: Being able to lazily initiate loading with a flow was a nice side effect of use of `produceState`

PR #874, opened 2022-08-24
MVI branch

tunjid:
An experimental branch for MVI style state production.

It offers feature parity with the existing branch and tries to establish some core MVI state production concepts:

Idiomatically functional: All inputs are eventually reduced into the UI state
No side effects: all changes to state must be modeled in the stream
Restricted UI state visibility: The state production pipeline cannot read the UI state until is ready to make changes to it
Well defined inputs: All information required to process an action must be present at the input of the action


MeandNi:
why r you delete mvi and mvvv branch?😂

PR #878, opened 2022-09-09
Makes effects in composables more declarative

manuelvicnt:


app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskScreen.kt
```diff
@@ -84,9 +85,10 @@ fun AddEditTaskScreen(
         )
 
         // Check if the task is saved and call onTaskUpdate event
-        LaunchedEffect(uiState.isTaskSaved) {
-            if (uiState.isTaskSaved) {
-                onTaskUpdate()
+        if (uiState.isTaskSaved) {
```
tunjid, line 88: Would this run on every recomposition, regardless of if `uiState.isTaskSaved` changed or not?

johnjohndoe: Maybe this is answered here: https://stackoverflow.com/questions/69085027/difference-between-remember-and-rememberupdatedstate-in-jetpack-compose

manuelvicnt: Yeah. This if statement will be check on every recomposition

adamp, line 88: This is using recomposition as an observer of state change events to apply other changes to app state. Avoid this.  Observe the state changes inside the `LaunchedEffect` instead so that it doesn't take one recomposition pass to apply `currentOnTaskUpdate` and then another to apply changes caused by that call.

adamp (changes requested):

Issue #861, opened 2022-06-22
"Usecases" link is not linked properly in the README.md


sumansahoo1:
![Screenshot from 2022-06-22 19-38-55](https://user-images.githubusercontent.com/75962762/175050358-fd196d63-b582-4f3d-be87-c740f2a10b38.png)
![Screenshot from 2022-06-22 19-39-11](https://user-images.githubusercontent.com/75962762/175050365-e755aa49-ab9d-4f0e-a128-5d2cac7f1c79.png)

VaradGupta23:
Solve This Error:
To fix this, you need to change the URL in the README.md to correctly point to the usecases directory.

Original (Broken) Link in README.md:

Markdown

[usecases](https://github.com/android/architecture-samples/tree/main/usecases)README.md:

You should change blob to tree in the URL to indicate you are linking to a directory.

Markdown

[usecases](https://github.com/android/architecture-samples/tree/main/usecases)
Explanation:
blob: Used for linking directly to a file in a specific branch/commit.

tree: Used for linking to a directory (folder) in a specific branch/commit, allowing you to browse its contents.

By changing blob to tree, the link will now correctly navigate to the usecases directory within the main branch of the android/architecture-samples repository on GitHub.

Steps to fix:

Open the README.md file in your text editor.

Find the line containing the broken "usecases" link.

Modify the URL within the parentheses () from .../blob/main/usecases to .../tree/main/usecases.

Save the README.md file.

If this is a Git repository, commit the change and push it to your remote (e.g., GitHub) for the fix to be visible online.

PR #880, opened 2022-10-01
Update README.md

suman-somu:
solves issue #861


jimbroze (changes requested):
That link no longer works

PR #882, opened 2022-11-01
Add packaging info to be picked up by the samples indexer

tunjid:


README.md
```diff
@@ -32,7 +32,7 @@ This project hosts each sample app in separate repository branches. For more inf
 
 ## Screenshots
 
-<img src="screenshots/todoapp.gif" alt="Screenshot">
```
mmoczkowski: Can we safely remove the old screenshot now?

tunjid: I decided to leave it in because they're still currently indexed and served to pages, I'll do a follow up CL to remove them once the new screenshots have been indexed. A cron job runs daily for this, so conservative TTL is 36 hrs.

PR #906, opened 2023-02-22
Replace `Result` with exception handling

dturner:
Here's what I've done: 

- Fixed a couple of unit tests where the behaviour was broken, specifically in `DefaultTaskRepositoryTest`
- Replaced all instances of `Result.Error` with `throw Exception`
- Added `catch` blocks to flow collection and added a couple of user error messages to indicate to them what went wrong
- Removed the `Result` class
- Changed the return type of `getTask` from `Result<Task>` to `Task?`, returning `null` indicates that the `Task` doesn't exist
- Updated unit tests to verify new behaviour


johnjohndoe:
Can you please explain **why** you made the changes?

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/DefaultTasksRepository.kt
```diff
@@ -110,9 +106,7 @@ class DefaultTasksRepository(
 
     override suspend fun completeTask(taskId: String) {
         withContext(ioDispatcher) {
-            (getTaskWithId(taskId) as? Success)?.let { it ->
-                completeTask(it.data)
-            }
+            getTaskWithId(taskId)?.run { completeTask(this) }
```
manuelvicnt, line 109: Nit: `apply` or `let` instead of `run`? You're modifying the task, it's not a random side-effect

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/DefaultTasksRepository.kt
```diff
@@ -125,9 +119,7 @@ class DefaultTasksRepository(
 
     override suspend fun activateTask(taskId: String) {
         withContext(ioDispatcher) {
-            (getTaskWithId(taskId) as? Success)?.let { it ->
-                activateTask(it.data)
-            }
+            getTaskWithId(taskId)?.run { activateTask(this) }
```
manuelvicnt, line 122: Ditto

app/src/main/java/com/example/android/architecture/blueprints/todoapp/statistics/StatisticsViewModel.kt
```diff
@@ -67,24 +65,19 @@ class StatisticsViewModel @Inject constructor(
...
-                        StatisticsUiState(
-                            isEmpty = result.data.isEmpty(),
-                            activeTasksPercent = stats.activeTasksPercent,
-                            completedTasksPercent = stats.completedTasksPercent,
-                            isLoading = false
-                        )
-                    }
-                    else -> StatisticsUiState(isLoading = false)
-                }
+                val stats = getActiveAndCompletedStats(taskLoad.data)
```
manuelvicnt, line 74: Could this throw an exception? If so, we need to handle it

app/src/main/java/com/example/android/architecture/blueprints/todoapp/taskdetail/TaskDetailViewModel.kt
```diff
@@ -63,6 +62,10 @@ class TaskDetailViewModel @Inject constructor(
     private val _isTaskDeleted = MutableStateFlow(false)
     private val _taskAsync = tasksRepository.getTaskStream(taskId)
         .map { handleResult(it) }
+        .catch {
+            showSnackbarMessage(R.string.loading_task_error) // TODO: Move to separate function?
```
manuelvicnt, line 66: Do we need to show this snackbar message? The function that handles the async response should be the one acting when that happens

app/src/main/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksViewModel.kt
```diff
@@ -72,6 +71,10 @@ class TasksViewModel @Inject constructor(
         combine(tasksRepository.getTasksStream(), _savedFilterType) { tasks, type ->
             filterTasks(tasks, type)
         }
+            .catch {
+                showSnackbarMessage(R.string.loading_tasks_error)
```
manuelvicnt, line 75: Ditto. This looks like a smell. `catch` shouldn't create a side effect that updates the ui state. It should notify the collector appropriately. Instead of an `Async.Success`, it should probably be an `Async.Error` if we really care about that error message.

It doesn't seem like that snackbar was shown before when the Result was an error, right?

app/src/main/java/com/example/android/architecture/blueprints/todoapp/statistics/StatisticsViewModel.kt
```diff
@@ -53,7 +51,7 @@ class StatisticsViewModel @Inject constructor(
     val uiState: StateFlow<StatisticsUiState> =
         tasksRepository.getTasksStream()
             .map { Async.Success(it) }
-            .onStart<Async<Result<List<Task>>>> { emit(Async.Loading) }
+            .onStart<Async<List<Task>>> { emit(Async.Loading) }
```
tunjid, line 54: Do we want to start with `Async.Loading` each time the stream is resubscribed to? It will case the same behavior from NiA of losing scroll state.

app/src/main/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksViewModel.kt
dturner: > It doesn't seem like that snackbar was shown before when the Result was an error, right?

The snackbar was shown previously when it was `Result.Error`: https://github.com/android/architecture-samples/pull/906/files#diff-631edb6efa01767b52856a0e498dc3987e8a2f0e8ba02837377dfbc5684af0cfL153

I'll create an `Async.Error` to handle this.

app/src/main/java/com/example/android/architecture/blueprints/todoapp/taskdetail/TaskDetailViewModel.kt
dturner: In a real app you might be on the task detail screen when a network sync fails. The message could indicate that the current task might not be up to date (bit tenuous but I was looking for places to show exception handling).

Have refactored to emit Async.Error`.

app/src/main/java/com/example/android/architecture/blueprints/todoapp/statistics/StatisticsViewModel.kt
dturner: This method can't, but `getTasksStream` can. I'll catch that, but this screen has no current provision for showing user messages so may have to just show an empty Ui state.

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/DefaultTasksRepository.kt
dturner: Unfortunately `also` doesn't work because there are 2 `completeTask` methods. Will keep it simple and change to `let`.

johnjohndoe:
Thank you for the explanation. I do not have insights into the codebase but I was asking myself why would a task but exist in the first place, motivating you to indicate the lack with `null`.?

tunjid (approved):

app/src/sharedTest/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskScreenTest.kt
```diff
@@ -109,7 +108,7 @@ class AddEditTaskScreenTest {
             .performClick()
 
         // THEN - Verify that the repository saved the task
-        val tasks = (repository.getTasksBlocking(true) as Result.Success).data
+        val tasks = repository.getTasksBlocking(true)
```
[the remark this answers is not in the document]
dturner: :bufo-happy:

app/src/main/java/com/example/android/architecture/blueprints/todoapp/taskdetail/TaskDetailViewModel.kt
```diff
@@ -120,11 +121,11 @@ class TaskDetailViewModel @Inject constructor(
...
 
-    private fun handleResult(tasksResult: Result<Task>): Async<Task?> =
-        if (tasksResult is Success) {
-            Async.Success(tasksResult.data)
+    private fun handleResult(tasksResult: Task?): Async<Task?> =
+        if (tasksResult != null) {
+            Async.Success(tasksResult)
         } else {
-            showSnackbarMessage(R.string.loading_tasks_error)
+            showSnackbarMessage(R.string.task_not_found)
```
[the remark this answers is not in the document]
dturner: Good spot, thank you. Fixed.

PR #907, opened 2023-02-27
Remove withContext, refactor duplicate methods

dturner:
Here's what I've done and why: 

- Removed all instances of `withContext`. `withContext` requires the caller to know something about the execution of the enclosed method(s) (e.g. it is cpu or disk bound). In this case, these calls are redundant because the only I/O bound operations are handled by Room, which will already move those operations onto the I/O thread, so we don't need to be concerned with it.
- Removed the `completeTask(task: Task)` and `activateTask(task: Task)` methods in favour of using the existing `completeTask(taskId: String)` and `activateTask(taskId: String)`. This creates a smaller API surface which is more loosely coupled. 
- Refactored `Task` to be immutable. Having a mutable model is an anti-pattern as it allows callers to modify the model, which violates SSOT.

### DIFF
--- a/app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/DefaultTasksRepository.kt
+++ b/app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/DefaultTasksRepository.kt
@@ -17,12 +17,9 @@
 package com.example.android.architecture.blueprints.todoapp.data.source
 
 import com.example.android.architecture.blueprints.todoapp.data.Task
-import kotlinx.coroutines.CoroutineDispatcher
-import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.withContext
 
 /**
  * Default implementation of [TasksRepository]. Single entry point for managing tasks' data.
@@ -30,7 +27,6 @@ import kotlinx.coroutines.withContext
 class DefaultTasksRepository(
     private val tasksRemoteDataSource: TasksDataSource,
     private val tasksLocalDataSource: TasksDataSource,
-    private val ioDispatcher: CoroutineDispatcher = Dispatchers.IO
 ) : TasksRepository {
 
     override suspend fun getTasks(forceUpdate: Boolean): List<Task> {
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/DefaultTasksRepository.kt
```diff
@@ -97,29 +93,17 @@ class DefaultTasksRepository(
...
-        withContext(ioDispatcher) {
-            getTaskWithId(taskId)?.let { completeTask(it) }
-        }
-    }
-
-    override suspend fun activateTask(task: Task) = withContext<Unit>(ioDispatcher) {
         coroutineScope {
-            launch { tasksRemoteDataSource.activateTask(task) }
-            launch { tasksLocalDataSource.activateTask(task) }
+            launch { tasksRemoteDataSource.completeTask(taskId) }
```
tunjid, line 98: Unrelated to your change, but the relationship between the remote and local data sources are a bit unclear here.
Since both are launched run in separate coroutines, either can fail and the failures are independent of each other.

If failing in the remote should not update the local, lets call them both in the same coroutine. If they are truly independent and should run simultaneously, lets be explicit and use [`async`](https://kotlinlang.org/docs/composing-suspending-functions.html#concurrent-using-async)

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/DefaultTasksRepository.kt
```diff
@@ -131,11 +115,9 @@ class DefaultTasksRepository(
     }
 
     override suspend fun deleteAllTasks() {
```
tunjid, line 117: If the function is already marked `suspend`, I wonder if we still need the `coroutineScope`, and to explicitly `launch` again in it.

dturner: Yeah, I'm not sure why this was implemented using `coroutineScope`. Presumably so that if one job fails then the other one is immediately cancelled, but that's not really what we want. 

I will add offline-first lazy writes in a separate PR which will resolve this issue.

manuelvicnt: It was implemented using `coroutineScope` so that you could `launch` multiple coroutines in parallel

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/remote/TasksRemoteDataSource.kt
```diff
@@ -80,22 +80,16 @@ object TasksRemoteDataSource : TasksDataSource {
...
-
     override suspend fun completeTask(taskId: String) {
-        // Not required for the remote data source
-    }
-
-    override suspend fun activateTask(task: Task) {
-        val activeTask = Task(task.title, task.description, false, task.id)
-        TASKS_SERVICE_DATA[task.id] = activeTask
+        TASKS_SERVICE_DATA[taskId]?.apply {
+            isCompleted = true
```
manuelvicnt, line 85: Having a model with mutable variables looks like a smell as changes to them won't cause new emissions to collectors.

We should change the `var`s in `Task` for `val`s unless there's a very important reason for not doing so

manuelvicnt: That could be a TODO for a new PR, but this change will need to be reverted in there

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/DefaultTasksRepository.kt
dturner: Gotcha, thanks.

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/remote/TasksRemoteDataSource.kt
dturner: Agreed. This is a problem with the existing model, but exacerbated by my change. I will change in this PR.

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/DefaultTasksRepository.kt
dturner: Will address in separate PR

tunjid (approved):

manuelvicnt (approved):

PR #908, opened 2023-03-03
Add data models, refactor local and network data sources

dturner:
Here's what I've done and why: 

**Add internal data models**
These are: 

- `NetworkTask` represents a task obtained from the network
- `LocalTask` represents a task stored in a local Room database

I've also added mapping functions to map between the two new models and the original `Task` model, which is exposed by the data layer to the UI layer. This is to more accurately mirror what would happen in a real app, where data obtained from a particular source needs to be mapped into a format which can be used by the rest of the app.   

They are in `ModelMappingExt.kt`, and could be converted to mapping classes if this is better practice.

**Local data source is now a Data Access Object**
Having both the local and network data sources implement the same interface wasn't realistic (they wouldn't have the same CRUD methods). The local data source is now a DAO. 

**Task creation and updates are now handled by the task repository**
I've moved the responsibility for creating and updating tasks into the task repo (with `createTask` and `updateTask` methods). Previously the UI layer (or any other layer) could create or update a `Task`, then the repo would save it. This violates SSOT because it's possible to have tasks with UUIDs in memory which the repository hasn't ever seen. This way the UUID creation is encapsulated in the repository, although it does mean the `create` and `update` methods need updating each time a new editable field is added to `Task`. Feedback welcome on this. 

**Tidied a few other things up**
- Removed `FakeFailingTasksRemoteDataSource.kt` - it's not used anywhere and I got tired of having build failures every time I changed the data source interface which it implements
- Removed the `repo.*blocking` convenience methods - tests have been updated to use `runTest` instead


app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/ModelMappingExt.kt
```diff
@@ -0,0 +1,95 @@
...
+ * `asNetworkModel`.
+ *
+ * - LocalTask: Internal model used to represent a task stored locally in a database. Obtained
+ * using `asLocalModel`.
+ *
+ */
+
+// External to local
+fun Task.asLocalModel(): LocalTask {
+    return LocalTask(
```
tunjid, line 39: For my curiosity, some methods here use use the expression body, and some use the block body. Is there a reason you prefer one or the other?

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/ModelMappingExt.kt
```diff
@@ -0,0 +1,95 @@
...
+        id = id,
+        title = title,
+        description = description,
+        isCompleted = isCompleted,
+    )
+}
+
+fun List<Task>.asLocalModels() = map { it.asLocalModel() }
```
tunjid, line 47: Maybe a method reference here? Maybe more trouble than its worth.

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/ModelMappingExt.kt
```diff
@@ -0,0 +1,95 @@
...
+    return LocalTask(
+        id = id,
+        title = title,
+        description = shortDescription,
+        isCompleted = (status == TaskStatus.COMPLETE),
+    )
+}
+
+@JvmName("networkTasksAsTaskEntities")
+fun List<NetworkTask>.asTaskEntities() = map { it.asTaskEntity() }
```
tunjid, line 73: Might want to add the annotation is to workaround type erasure on the JVM

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/DefaultTasksRepository.kt
```diff
@@ -83,48 +122,41 @@ class DefaultTasksRepository(
...
-            launch { tasksLocalDataSource.saveTask(task) }
-        }
+        return tasksDao.getTaskById(taskId)?.asExternalModel()
     }
 
     override suspend fun completeTask(taskId: String) {
         coroutineScope {
-            launch { tasksRemoteDataSource.completeTask(taskId) }
-            launch { tasksLocalDataSource.completeTask(taskId) }
+            launch { tasksNetworkDataSource.completeTask(taskId) }
```
tunjid, line 130: The codependency dependency of the coroutines launched here are a bit hard to reconcile. If any coroutine fails, while the other is running, the other one is cancelled. However either can succeed first, and the other can fail next.

I think it'd be a bit clearer if both must succeed, or if one fails, it doesn't force the other to be cancelled if it is still running.

tunjid (commented):
LGTM.

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/ModelMappingExt.kt
dturner: Nope :) Refactored to use expressions wherever possible.

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/ModelMappingExt.kt
dturner: Refactored to use method references.

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/ModelMappingExt.kt
dturner: Added

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/DefaultTasksRepository.kt
```diff
@@ -17,58 +17,97 @@
...
-    private val tasksRemoteDataSource: TasksDataSource,
-    private val tasksLocalDataSource: TasksDataSource,
+    private val tasksNetworkDataSource: NetworkDataSource,
+    private val tasksDao: TasksDao,
 ) : TasksRepository {
 
+    override suspend fun createTask(title: String, description: String): Task {
+        val task = Task(title = title, description = description)
+
+        coroutineScope {
```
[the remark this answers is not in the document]
dturner: I've deliberately avoided refactoring this code as it makes the (already large) PR enormous. I'd like to implement an example of offline-first using the methodology from Now in Android, but will do this in a different PR (if the team agrees).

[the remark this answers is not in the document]
tunjid: ohh gotcha

dturner:
Thanks @tunjid. Did you see my comments about the independent coroutines? This is the existing functionality - although it looks like new code it's actually exactly the same as how it works currently. I will address this in a separate PR if/when we redo the network data source and sync algorithm.

manuelvicnt (commented):
Overall LGTM

mjobayeraraf:
Hi. Who are you

PR #912, opened 2023-03-09
Add modules for data layer and test dependencies

### LINKED ISSUE #910: Project uses outdated approach to sharing dependencies between unit and instrumented tests
Using a `sharedTest` folder is no longer the recommended approach to sharing code between unit and instrumented tests. It's also pretty confusing in the Project explorer -> Android view to see where to add a new test, since the `sharedTest` folder just shows up as a duplicate "test" folder: 

image

Suggest implementing the recommendation from https://b.corp.google.com/issues/232420188#comment19 which is to create a module containing shared test code. This results in a clearer structure: 

image

dturner:
Here's what I've done and why.

I've created 3 new modules and refactored the existing code into them: 

- `shared-test` contains any shared test dependencies, such as `MainCoroutineRule`
- `data` contains the data layer (repositories, models and data sources), and any associated tests
- `data-test` contains test dependencies and fakes for the data layer, such as `FakeRepository`. It also contains the hilt modules which inject the test doubles during instrumented tests, such as `RepositoryTestModule`. 

Why do this? 
- Using a shared folder is not the recommended way of doing things any more (fixes #910)
- Although there is some extra complexity introduced by the modules, it's (imo) clearer where the tests reside i.e. `:data` has its unit tests in `test` and instrumented tests in `androidTest` - they're not hiding away inside the `sharedTest` folder.   

Note to reviewers
- `:data` is dependent on `:data-test` (because it provides the test dependencies), but `:data-test` is also dependent on `:data` because the Hilt modules depend on the interfaces defined in the data layer (e.g `RepositoryTestModule` depends on `TasksRepository`. This creates a circular dependency - is this ok? Is there a better way of achieving this?


tunjid (approved):
LGTM!

PR #913, opened 2023-03-10
Refactor tests to use assertEquals

dturner:
I've refactored `TasksDaoTest` to: 

- use `assertEquals` because it's easier to read and removes the dependency on Hamcrest
- instantiate the database only once, then clear it before each test - should result in slightly faster test execution

Also did a very minor tidy up to `AddEditTaskScreenTest` to swap expected and actual values as they were the wrong way round.

### DIFF
--- a/app/src/androidTest/java/com/example/android/architecture/blueprints/todoapp/data/source/local/TasksDaoTest.kt
+++ b/app/src/androidTest/java/com/example/android/architecture/blueprints/todoapp/data/source/local/TasksDaoTest.kt
@@ -22,12 +22,10 @@ import androidx.test.core.app.ApplicationProvider.getApplicationContext
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.filters.SmallTest
 import com.example.android.architecture.blueprints.todoapp.MainCoroutineRule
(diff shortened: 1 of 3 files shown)


app/src/androidTest/java/com/example/android/architecture/blueprints/todoapp/data/source/local/TasksDaoTest.kt
```diff
@@ -38,7 +36,12 @@ import org.junit.runner.RunWith
 @SmallTest
 class TasksDaoTest {
 
-    private lateinit var database: ToDoDatabase
+    // using an in-memory database because the information stored here disappears when the
+    // process is killed
+    private val database = Room.inMemoryDatabaseBuilder(
```
tunjid, line 41: We probs still want to do this in `@Before` and `@After` so results of prior tests in the same test suite don't affect new ones

dturner: How would tests affect previous tests if we're calling `clearAllTables` @Before each test?

JoseAlcerreca (approved):

PR #915, opened 2023-03-13
Simplify the network data source

dturner:
tl;dr the local data source is now the source of truth, the network data source has only 2 methods which load and save all data. 

One of the goals of this sample is to demonstrate how to deal with _different_ data sources but the `NetworkDataSource` interface was very similar to the local `TasksDao` interface. 

For this reason, I have simplified the `NetworkDataSource` interface to just 2 methods `loadTasks` and `saveTasks`, which load and save _all_ tasks respectively.

I've also changed the way that data synchronisation works inside `DefaultTasksRepository`: 

- When reading tasks they are obtained locally, unless `forceRefresh` is specified, in which case they are first obtained from the network data source
- When writing information they are written locally first, then written to the network data source

This is not a robust sync implementation. In a production app, if the remote write failed and a force refresh was performed the local data would be overwritten. But I think it's better than what we had before which was essentially random behaviour.

### DIFF
--- a/app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/DefaultTasksRepository.kt
+++ b/app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/DefaultTasksRepository.kt
@@ -81,26 +81,6 @@ class DefaultTasksRepository(
         loadTasksFromNetwork()
     }
 
-    private suspend fun loadTasksFromNetwork() {
-        withContext(coroutineDispatcher) {
-            val remoteTasks = tasksNetworkDataSource.loadTasks()
-
-            // Real apps might want to do a proper sync, deleting, modifying or adding each task.
-            tasksDao.deleteTasks()
-            remoteTasks.forEach { task ->
-                tasksDao.insertTask(task.toTaskEntity())
-            }
-        }
-    }
-
-    private suspend fun saveTasksToNetwork() {
-        withContext(coroutineDispatcher) {
-            // Real apps may want to use a proper sync strategy here to avoid data conflicts.
-            val localTasks = tasksDao.getTasks()
-            tasksNetworkDataSource.saveTasks(localTasks.toNetworkModels())
-        }
-    }
-
     override fun getTaskStream(taskId: String): Flow<Task?> {
         return tasksDao.observeTaskById(taskId).map { it.toExternalModel() }
     }
@@ -141,5 +121,25 @@ class DefaultTasksRepository(
     override suspend fun deleteTask(taskId: String) {
         tasksDao.deleteTaskById(taskId)
         saveTasksToNetwork()
+    }
+
+    private suspend fun loadTasksFromNetwork() {
+        withContext(coroutineDispatcher) {
+            val remoteTasks = tasksNetworkDataSource.loadTasks()
+
+            // Real apps might want to do a proper sync, deleting, modifying or adding each task.
+            tasksDao.deleteTasks()
+            remoteTasks.forEach { task ->
+                tasksDao.insertTask(task.toTaskEntity())
+            }
+        }
+    }
+
+    private suspend fun saveTasksToNetwork() {
+        withContext(coroutineDispatcher) {
+            // Real apps may want to use a proper sync strategy here to avoid data conflicts.
+            val localTasks = tasksDao.getTasks()
+            tasksNetworkDataSource.saveTasks(localTasks.toNetworkModels())
+        }
     }
 }
--- a/app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/DefaultTasksRepository.kt
+++ b/app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/DefaultTasksRepository.kt
@@ -129,16 +129,12 @@ class DefaultTasksRepository(
     }
 
     override suspend fun clearCompletedTasks() {
-        withContext(coroutineDispatcher) {
-            tasksDao.deleteCompletedTasks()
-        }
+        tasksDao.deleteCompletedTasks()
         saveTasksToNetwork()
     }
 
     override suspend fun deleteAllTasks() {
-        withContext(coroutineDispatcher) {
-            tasksDao.deleteTasks()
-        }
+        tasksDao.deleteTasks()
         saveTasksToNetwork()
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/DefaultTasksRepository.kt
```diff
@@ -95,68 +83,51 @@ class DefaultTasksRepository(
         }
     }
 
-    override fun getTaskStream(taskId: String): Flow<Task?> {
-        return tasksDao.observeTaskById(taskId).map { it.toExternalModel() }
+    private suspend fun sendTasksToNetworkDataSource() {
```
manuelvicnt, line 86: `sendTasksToNetworkDataSource` sounds too prescriptive with what the network is doing. Maybe we should choose a more generic name? We shouldn't know the network data source impl details.

Maybe something something shorter and more generic like `syncNetworkTasks` would do

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/DefaultTasksRepository.kt
```diff
@@ -95,68 +83,51 @@ class DefaultTasksRepository(
...
+        // Real apps may want to use a proper sync strategy here to avoid data conflicts.
+        val localTasks = tasksDao.getTasks()
+        tasksNetworkDataSource.saveTasks(localTasks.toNetworkModels())
```
manuelvicnt, line 89: Same comment applies to all places where we do mapping.

We don't know how expensive retrieving from the local data source and mapping the models is. We should move the execution to the `defaultDispatcher`. We could be talking about hundreds / thousands of tasks and that could block the calling thread.

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/DefaultTasksRepository.kt
dturner: I agree this name is too verbose, however, the direction of data travel and where it's going to is important. For this reason `sync` is too generic. How about just `sendTasksToNetwork`?

dturner: btw I chose this name to match the existing counterpart method `updateTasksFromNetworkDataSource`, I could also rename this to `loadTasksFromNetwork`.

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/source/DefaultTasksRepository.kt
dturner: Good point. I've moved execution to `Dispatchers.Default` for all *bulk* network, local db and mapping operations. Do you think it's worth doing this for single operations as well, like `getTask`? To me that seems like overkill.

dturner: Just to note that only bulk network and mapping operations need `withContext`, Room will take care of using the correct dispatcher for local database operations.

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/DefaultTasksRepository.kt
```diff
@@ -18,141 +18,128 @@ package com.example.android.architecture.blueprints.todoapp.data
...
     }
 
     override suspend fun refreshTask(taskId: String) {
-        updateTaskFromRemoteDataSource(taskId)
+        loadTasksFromNetwork()
     }
 
-    private suspend fun updateTasksFromRemoteDataSource() {
-        val remoteTasks = tasksNetworkDataSource.loadTasks()
+    private suspend fun loadTasksFromNetwork() {
```
manuelvicnt, line 84: Nit: move private functions below public overriden ones

manuelvicnt (approved):

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/DefaultTasksRepository.kt
dturner: Done

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/DefaultTasksRepository.kt
```diff
@@ -18,141 +18,128 @@ package com.example.android.architecture.blueprints.todoapp.data
...
     override suspend fun refreshTasks() {
-        updateTasksFromRemoteDataSource()
+        loadTasksFromNetwork()
     }
 
     override fun getTasksStream(): Flow<List<Task>> {
         return tasksDao.observeTasks().map { tasks ->
-            tasks.map { task ->
-                task.toExternalModel()
+            withContext(coroutineDispatcher) {
```
tunjid, line 74: Why do we need to change the context? `tasks.toExternalModels()` seems synchronous

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/DefaultTasksRepository.kt
```diff
@@ -18,141 +18,128 @@ package com.example.android.architecture.blueprints.todoapp.data
...
+            tasksDao.deleteTasks()
+            remoteTasks.forEach { task ->
+                tasksDao.insertTask(task.toTaskEntity())
+            }
+        }
+    }
+
+    private suspend fun saveTasksToNetwork() {
+        withContext(coroutineDispatcher) {
+            // Real apps may want to use a proper sync strategy here to avoid data conflicts.
```
tunjid, line 140: May also want to add that this may be work best deferred to `WorkManager`.

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/DefaultTasksRepository.kt
dturner: For exactly that reason, it is synchronous and could therefore block the calling thread indefinitely if it was, say, mapping a million tasks

PR #926, opened 2023-03-23
Add application scope for fire-and-forget jobs

dturner:
What have I done and why?

- Added a coroutine scope to `DefaultTaskRepository` so that it can launch the job which writes local to network data asynchronously. This fixes an issue in the current app where when creating or updating any task, the UI would pause for 2 seconds waiting for the network data source (which has a simulated 2 second network latency).

For the main app, an application scope is used. This is defined using the `ApplicationScope` qualifier. 
For tests, `TestScope` is used. 

- Refactored the hilt configuration to use `@Binds` rather than `@Provides` methods. This (in my mind anyway) makes things simpler to understand - `@Provides` should be reserved for objects which there is no other way to create (such as those from Room).  

- Updated `DefaultTaskRepositoryTest` to use `StandardTestDispatcher` rather than a hardcoded `Dispatchers.Main` when creating the `DefaultTaskRepository`

- Renamed `DefaultTaskRepository`'s parameters for the data sources to make it more obvious what they are.  

Note: This change also includes https://github.com/android/architecture-samples/pull/928


app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/DefaultTaskRepository.kt
```diff
@@ -18,25 +18,32 @@ package com.example.android.architecture.blueprints.todoapp.data
...
+ * @param fireAndForgetScope - The coroutine scope used for fire-and-forget jobs, such as sending
+ * data to the network.
  */
-class DefaultTaskRepository(
+@Singleton
+class DefaultTaskRepository @Inject constructor(
     private val tasksNetworkDataSource: NetworkDataSource,
     private val taskDao: TaskDao,
-    private val coroutineDispatcher: CoroutineDispatcher = Dispatchers.Default
+    private val expensiveWorkDispatcher: CoroutineDispatcher,
```
manuelvicnt, line 45:
```suggestion
    private val defaultDispatcher: CoroutineDispatcher,
```

It's ok to use `default` here as people know it's for expensive work operations

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/DefaultTaskRepository.kt
```diff
@@ -18,25 +18,32 @@ package com.example.android.architecture.blueprints.todoapp.data
...
+    private val fireAndForgetScope: CoroutineScope,
```
manuelvicnt, line 46:
```suggestion
    private val scope: CoroutineScope,
```

All scopes are fire and forget. It's fine to say `externalScope` or simply `scope`

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/DefaultTaskRepository.kt
```diff
@@ -136,15 +143,15 @@ class DefaultTaskRepository(
...
             val remoteTasks = tasksNetworkDataSource.loadTasks()
             taskDao.deleteAll()
             taskDao.upsertAll(remoteTasks.toLocal())
         }
     }
 
-    private suspend fun saveTasksToNetwork() {
-        withContext(coroutineDispatcher) {
+    private fun saveTasksToNetwork() {
+        fireAndForgetScope.launch {
```
manuelvicnt, line 154: As discussing in Chat... what about errors here?

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/DefaultTaskRepository.kt
```diff
@@ -18,26 +18,33 @@ package com.example.android.architecture.blueprints.todoapp.data
...
 
 /**
  * Default implementation of [TaskRepository]. Single entry point for managing tasks' data.
  *
  * @param tasksNetworkDataSource - The network data source
  * @param taskDao - The local data source
- * @param coroutineDispatcher - The dispatcher to be used for long running or complex operations,
- * such as network calls or mapping many models. This is important to avoid blocking the calling
- * thread.
+ * @param expensiveWorkDispatcher - The dispatcher to be used for long running or complex
```
manuelvicnt, line 36: Nit: change KDocs as well

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/DefaultTaskRepository.kt
```diff
@@ -70,7 +77,7 @@ class DefaultTaskRepository(
         if (forceUpdate) {
             loadTasksFromNetwork()
         }
-        return withContext(coroutineDispatcher) {
+        return withContext(expensiveWorkDispatcher) {
```
manuelvicnt, line 80: These've been renamed. You need to update them all

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/DefaultTaskRepository.kt
manuelvicnt: Let's add a comment and that should be enough

app/src/main/java/com/example/android/architecture/blueprints/todoapp/di/DataModules.kt
```diff
@@ -44,8 +47,15 @@ object RepositoryModule {
     fun provideTaskRepository(
         @NetworkTaskDataSource remoteDataSource: NetworkDataSource,
         database: ToDoDatabase,
+        @DefaultDispatcher dispatcher: CoroutineDispatcher,
```
manuelvicnt, line 50: Nit: Update with the implementation that provides the `TaskDao` instead of the repo

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/DefaultTaskRepository.kt
dturner: Done

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/DefaultTaskRepository.kt
dturner: Done

app/src/main/java/com/example/android/architecture/blueprints/todoapp/di/DataModules.kt
dturner: I've replaced `provides` with `binds` for the `TaskRepository` which is quite a bit simpler.

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/DefaultTaskRepository.kt
```diff
@@ -103,60 +107,74 @@ class DefaultTaskRepository(
...
-    private suspend fun saveTasksToNetwork() {
-        withContext(coroutineDispatcher) {
-            val localTasks = taskDao.getAll()
-            tasksNetworkDataSource.saveTasks(localTasks.toNetwork())
+    /**
+     * Send the tasks from the local data source to the network data source
+     *
+     * Returns immediately after launching the job. Real apps may want to either suspend here until
+     * the operation is complete, or provide a mechanism for the result to be communicated back
+     * to other layers.
```
tunjid, line 172: This is a great example of deferred work, so the comment should explicitly say real apps may want to defer to `WorkManager` here. We implement a basic form deferred work by launching an independent coroutine with no exponential backoff. We also probs want to catch any exceptions for completeness sake.

tunjid (approved):

app/src/main/java/com/example/android/architecture/blueprints/todoapp/data/DefaultTaskRepository.kt
dturner: Comment updated.

manuelvicnt (approved):

PR #935, opened 2023-03-28
Minor tidy ups

dturner:
Here's what I've done: 

- Added `suspend` to `upsertAll` (it was missed in a previous PR)
- Move the use of `withContext` to only wrap the mapping functions 
- Remove `MainCoroutineRule` from both `TaskDaoTest` and `DefaultTaskRepositoryTest`. In the latter, a `testScope` and `testDispatcher` are used to guarantee deterministic behaviour
- Remove pluralization `tasksRepository` to `taskRepository` in line with other recent changes

### DIFF
--- a/app/src/androidTest/java/com/example/android/architecture/blueprints/todoapp/data/source/local/TaskDaoTest.kt
+++ b/app/src/androidTest/java/com/example/android/architecture/blueprints/todoapp/data/source/local/TaskDaoTest.kt
@@ -35,15 +35,16 @@ class TaskDaoTest {
 
     // using an in-memory database because the information stored here disappears when the
     // process is killed
-    private val database = Room.inMemoryDatabaseBuilder(
-        getApplicationContext(),
-        ToDoDatabase::class.java
-    ).allowMainThreadQueries().build()
+    private lateinit var database: ToDoDatabase
 
-    // Ensure that we use an empty database for each test.
+    // Ensure that we use a new database for each test.
     @Before
-    fun initDb() = database.clearAllTables()
-
+    fun initDb() {
+        database = Room.inMemoryDatabaseBuilder(
+            getApplicationContext(),
+            ToDoDatabase::class.java
+        ).allowMainThreadQueries().build()
+    }
     @Test
     fun insertTaskAndGetById() = runTest {
         // GIVEN - insert a task
--- a/app/src/androidTest/java/com/example/android/architecture/blueprints/todoapp/data/source/local/TaskDaoTest.kt
+++ b/app/src/androidTest/java/com/example/android/architecture/blueprints/todoapp/data/source/local/TaskDaoTest.kt
@@ -42,10 +42,6 @@ class TaskDaoTest {
         ToDoDatabase::class.java
     ).allowMainThreadQueries().build()
 
-    // Set the main coroutines dispatcher for unit testing.
-    @get:Rule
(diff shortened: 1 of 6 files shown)


app/src/androidTest/java/com/example/android/architecture/blueprints/todoapp/data/source/local/TaskDaoTest.kt
```diff
@@ -42,10 +40,6 @@ class TaskDaoTest {
         ToDoDatabase::class.java
     ).allowMainThreadQueries().build()
 
-    // Set the main coroutines dispatcher for unit testing.
-    @get:Rule
-    val mainCoroutineRule = MainCoroutineRule()
-
     // Ensure that we use an empty database for each test.
     @Before
     fun initDb() = database.clearAllTables()
```
manuelvicnt, line 45: This shouldn't be needed either, a new database will be created per test

app/src/test/java/com/example/android/architecture/blueprints/todoapp/data/DefaultTaskRepositoryTest.kt
```diff
@@ -101,19 +97,16 @@ class DefaultTaskRepositoryTest {
     @Test
     fun getTasks_requestsAllTasksFromRemoteDataSource() = runTest {
```
manuelvicnt, line 98: All `runTest`s in this file should use the `testScope` instead

app/src/androidTest/java/com/example/android/architecture/blueprints/todoapp/data/source/local/TaskDaoTest.kt
dturner: What magic causes the database to be re-created for each test? It's only defined once in the test class.

app/src/test/java/com/example/android/architecture/blueprints/todoapp/data/DefaultTaskRepositoryTest.kt
dturner: Done

app/src/androidTest/java/com/example/android/architecture/blueprints/todoapp/data/source/local/TaskDaoTest.kt
manuelvicnt: It might be the same instance in all tests but the `inMemoryDatabase` is cleared when the process is killed. If that happens after every test run, that's why we might see an empty database.

IMO, I think it's better to avoid these variables in the test class itself. The initialization should be done in the `@Before` methods. I'd have something like this:

```
class TaskDaoTest {
    lateinit var database: ToDoDatabase

    @Before
    fun init() {
        database = Room.inMemoryDatabase(...)
    }
}
```

My understanding is that if the same instance is reused across tests, that wouldn't make them elegible to be run in parallel. 

Maybe @JoseAlcerreca can chime in here as well

manuelvicnt (approved):

app/src/androidTest/java/com/example/android/architecture/blueprints/todoapp/data/source/local/TaskDaoTest.kt
dturner: Code updated to use your approach - it's safer than just clearing the tables.

PR #937, opened 2023-04-09
Migrate to KSP

AnirudhPudari:
What Changes are done?
- Migrated room to use ksp

What Changes are not done as per requirement ?
- Couldn't migrate hilt to use ksp as the support for both Dagger and Hilt is not yet available. Please find the list of supported libraries on Android with support for KSP . Please check list of libraries supported for KSP on Android https://kotlinlang.org/docs/ksp-overview.html#supported-libraries


tunjid:
@AnirudhPudari thanks so much for doing this! Will merge as soon as CI passes.

PR #951, opened 2023-05-19
Fix dependencies, compile with Java 19, added Quit action to ... menu

rssnow1:
I had to switch to using the Beta version of android  studio because of a Kotlin compiler mismatch with the Compose compiler versions.
Also added Quit action to the ... menu.


rssnow1:
switched to beta studio, added Quit action to menu.

PR #952, opened 2023-05-19
Added Quit to menu

rssnow1:
Added a Quit action to the ... menu to exit the app without having to press back.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -1,4 +1,3 @@
-import com.android.build.api.dsl.Packaging
 import org.gradle.api.JavaVersion.VERSION_17
 
 
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -1,3 +1,7 @@
+import com.android.build.api.dsl.Packaging
+import org.gradle.api.JavaVersion.VERSION_17
+
+
 /*
  * Copyright 2020 The Android Open Source Project
  *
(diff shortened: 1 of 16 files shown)


rssnow1 (commented):
With all of these changes, it builds with the beta android studio using oracle jdk 19

rssnow1:
The last change was to add Quit to the menu ...

app/src/main/java/com/example/android/architecture/blueprints/todoapp/TodoNavigation.kt
```diff
@@ -102,5 +105,11 @@ class TodoNavigationActions(private val navController: NavHostController) {
                 if (taskId != null) "$it?$TASK_ID_ARG=$taskId" else it
             }
         )
+
     }
+
+    fun onQuit() {
+        System.exit(0)
```
joezorry, line 112: Never do this in an android app, it can mess up with activities lifecycle.


Also in this instance, there is really no need.

app/build.gradle.kts
```diff
@@ -20,15 +23,18 @@ plugins {
...
 
 android {
-    compileSdk = libs.versions.compileSdk.get().toInt()
 
+    compileSdkPreview = "UpsideDownCake"
     defaultConfig {
         applicationId = "com.example.android.architecture.blueprints.main"
-        minSdk = libs.versions.minSdk.get().toInt()
-        targetSdk = libs.versions.targetSdk.get().toInt()
+        minSdk = 33
```
joezorry, line 36: This should still be the ones referenced from the libs file.

rssnow1: > This should still be the ones referenced from the libs file.
 and what is that file named?

app/src/main/java/com/example/android/architecture/blueprints/todoapp/TodoNavigation.kt
rssnow1: ok so on my phone (Pixel) I just figured out the gesture for Back is swipe up from bottom. 

I wasn't that familiar with the gestures. 

I am running the new android beta

rssnow1 (commented):
I see now that swiping up from the bottom of the screen is now the way to go back. 

I just didn't see a way before to get the app to leave the screen. 

I am using  the beta image on my pixel, I started using android some where back in Lollipop days lol.

PR #963, opened 2023-07-24
TextDecoration LineThrough-None

Askeri-Muhendis:
If we assign null value after assigning value to textDecoration, it will not work as expected. We need to create an instance over the companion object.


Askeri-Muhendis:
In the new versions, the android team has fixed it, but there is a problem because there is an old version in the project. you can follow here 
https://github.com/android/architecture-samples/pull/964

PR #964, opened 2023-07-24
Version Update

ibrahimethemsen:
Version Update & Null bug in textDecoration, versions solved together


metehanbolatt:
The situation you mentioned in the current version seems to have been resolved. Thanks

PR #967, opened 2023-08-12
Update gradle

MJJacobs01:
Updated the project to use the latest gradle version.
I have also updated all dependencies to the highest versions that will still past all the tests.

I have picked up the from Kotlin version 1.8.21 there is a reported issue with using ksp, so composeBom, compose compiler and Kotlin cannot be updated to more recent version currently.


MJJacobs01:
Good morning,

Is this pull request going to be accepted?

PR #978, opened 2023-10-03
Fixed: Issue - On uncheck checkbox (mean active the task) the strike through was not removing.

NalinMeher:


dturner:
Thanks for doing this 👍

PR #1011, opened 2024-09-24
Create Architecture sample

CoolDude691:


CoolDude691:
Explain architecture pattern

PR #1061, opened 2026-03-28
Fix: prevent repeated snackbar messages by migrating state to TasksViewModel

masoomyf:
This PR fix a bug where Snackbar messages (such as "Task Saved") would reappear when navigating back to the TasksScreen or during configuration changes.

The root cause was the manual management of navigation arguments within the UI layer. Passing the userMessage through the TasksScreen composable and attempting to reset it via a callback was never succeed, because it can't be update.


hoodbarbie727glee-sudo:
#todoapp-mvvm-databinding-live (d0250f9d76149d05a4ec1ce8c807d8ec8866124e)

